Issue #63, opened 2014-12-09
[Rest] add verbose filter to Controller/ClassContentController::getCollectionAction()


eric-chau:
verbose default value is __true__, this is the default and actual behavior. When verbose is setted to __false__, this method will only return the definition of classcontent; it means that you won't get elements or uid entries.

eric-chau:
Rework ClassContentController API this way:

- ``/rest/{version}/classcontent/{type}`` must return objects complete data
- ``/rest/{version}/classcontent/{type}?defintion`` must return objects definition, which means only unchanging attributes (maxentry, minentry, accept and properties)
- ``/rest/{version}/classcontent/{type}?concise`` must return objects specifics attributes like elements, uid, type.
- ``/rest/{version}/classcontent/{type}?info`` must return objects infos like created, modified, etc.

The idea is when you combine object's definition, concise and info you fallback to object full data (except to elements which won't be verbose as complete data version).

eric-chau:
Rework of ClassContentController API implies several things like deleting definition routes. To avoid this API regression we have to add a new route to get all classcontents' definitions: ``/rest/{version}/classcontent?definition``.

Also, should we allow ``/rest/{version}/classcontent`` to return something? It may be not enough restrictive.

ping @crouillon.

PR #109, opened 2014-12-11
[RFR] [ClassContent] #63 + add verbose filter to getAction() and getCollectionAction() of ClassContentController

eric-chau:
This option is passed as query parameter; it's necessary to lighten get classcontent action response;

The reduction is done by unsetting classcontent entry provided by classcontent definition (like maxentry, minentry, accept and properties)

I did some benchmarks:

- get article collection (26 items) with verbose true: response with over 20k lines
- get article collection (26 items) with verbose false: response with ~9k lines

It still heavy so we have to discuss about this with @crouillon.

### DIFF
--- a/ClassContent/AContent.php
+++ b/ClassContent/AContent.php
@@ -1016,9 +1016,9 @@ abstract class AContent implements IObjectIdentifiable, IRenderable, \JsonSerial
     /**
      * {@inheritdoc}
      */
-    public function jsonSerialize()
+    public function jsonSerialize($verbose = true)
     {
-        $datas = array(
+        $data = array(
             'uid'        => $this->_uid,
             'label'      => $this->_label,
             'type'       => str_replace(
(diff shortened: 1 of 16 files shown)


eric-chau:
Need to update ClassContentController:

- remove definition route
- update ``::getAction`` and ``::getCollectionAction``

ClassContent/AContent.php
```diff
@@ -47,6 +47,14 @@
     const CLASSCONTENT_BASE_NAMESPACE = 'BackBee\ClassContent\\';
```
mickaelandrieu: We should use ``::class`` [operator](http://php.net/manual/en/language.oop5.basic.php#language.oop5.basic.class.class). 

What is / will be the minimal PHP version to use BackBee ?

ClassContent/AContent.php
```diff
@@ -870,7 +877,7 @@ public function getParam($var = null, $type = null)
             array_unshift($pieces, $type);
         }
 
-        return $this->_getRecursivelyParam($this->_parameters[$var], $pieces);
+        return $this->getRecursivelyParam($this->_parameters[$var], $pieces);
```
mickaelandrieu: :+1:

eric-chau:
ping @backbee/core-team, I need to merge it very soonly. Thanks.

ClassContent/AContent.php
eric-chau: BackBee min version is 5.4.0. Anyway, it could not work cause ``::class`` operator returns ``BackBee\ClassContent\AContent`` when we want to get ``BackBee\ClassContent\``, that implies string treatment which is useless since we know what we want and it does not need to be dynamic.

eric-chau:
Everything is green, last ping @backbee/core-team before merge.

PR #136, opened 2014-12-19
[WIP] - Resolved #2 + Added missing use statments

### LINKED ISSUE #2: Missing use statement should be avoided
There is 2 issues in this report: 

* Some statments have been forgotten
* An autoloading issue on Doctrine annotations

I will fix this both 2 issues

mickaelandrieu:
First of all, notice this is a **work in progress**

As stated in #2, there is 2 sorts of "missing use"
* the statment is missing: we just have to add the correct statment
* the annotations from Doctrine, which are auto-loaded without need of ``use``

As far as I have investigated, this is not an ``AnnotationReader`` but a ``SimpleAnnotationReader`` which is used.

So, we need to add "every" missing statments. This will resolve the issue until we take a big review of our custom auto-loader...

Finaly, I've renamed ``Cache\DAO\Entity`` to ``Cache\DAO\CacheEntity`` because of ambiguous behaviors on ``@Entity`` annotation used in this class.

### DIFF
--- a/Logging/Logger.php
+++ b/Logging/Logger.php
@@ -119,9 +119,9 @@ class Logger extends DebugStack implements LoggerInterface, SQLLogger
             }
         }
 
-        $this->_setErrorHandler()
+        /*$this->_setErrorHandler()
              ->_setExceptionHandler()
-        ;
+        ;*/
     }
 
     public function __destruct()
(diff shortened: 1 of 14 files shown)


Logging/Logger.php
```diff
@@ -119,9 +119,9 @@ public function __construct(BBApplication $application = null, IAppender $append
             }
         }
 
-        $this->_setErrorHandler()
+        /*$this->_setErrorHandler()
              ->_setExceptionHandler()
-        ;
+        ;*/
```
mickaelandrieu: needed to be rebased as this part of code does'nt exists anymore.

Issue #158, opened 2014-12-31
[Logger] improve the way we define the logfile filepath


eric-chau:
Actually if you provide a filepath for the logfile and the folder does not exist, it will generate a wrong folder structure. Example:

Admit that the folder ``/var/www/backbee-standard/log`` does not exist and our logfile parameter is setted to ``/var/www/backbee-standard/log/backbee.log``. Cause the folder does not exist, Logger will create logfile into this path: ``/var/www/backbee-standard/var/www/backbee-standard/log/backbee.log``

It's mainly due to (BackBee\Logging\Logger:106):

```php
if (array_key_exists('logfile', $loggingConfig)) {
    if (false === realpath(dirname($loggingConfig['logfile']))) {
        $loggingConfig['logfile'] = $application->getBaseDir().DIRECTORY_SEPARATOR.$loggingConfig['logfile'];
    }
}
```

PR #161, opened 2015-01-05
[RFR] [Logging] #158 - removed validation logfile step from Logger

eric-chau:
This step should not be done in Logger but in Appender\File cause logfile is only used in case that requested appender is Appender\File.


mickaelandrieu:
:+1:

Issue #162, opened 2015-01-05
[Installer] update EntityFinder::getNamespace method


eric-chau:
This method built valid namespace for BackBee when it was located on project's root. But it does not do the job anymore since ``backbee/backbee`` moved into vendor folder.

PR #163, opened 2015-01-05
[WIP] [Installer] #162 update EntityFinder::getNamespace method

eric-chau:
This method built valid namespace for BackBee when it was located on project's root. But it does not do the job anymore since backbee/backbee moved into vendor folder.

### DIFF
--- a/Services/Local/Media.php
+++ b/Services/Local/Media.php
@@ -29,7 +29,6 @@ use BackBee\ClassContent\Element\image as elementImage;
 use BackBee\ClassContent\Element\text as elementText;
 use BackBee\ClassContent\Exception\ClassContentException;
 use BackBee\Services\Exception\ServicesException;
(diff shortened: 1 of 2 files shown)


Services/Local/Media.php
```diff
@@ -29,7 +29,6 @@
 use BackBee\ClassContent\Element\text as elementText;
 use BackBee\ClassContent\Exception\ClassContentException;
 use BackBee\Services\Exception\ServicesException;
-use BackBee\Util\Media;
```
mickaelandrieu: hmm, why ?

I like to see what classes are used in class header.

mickaelandrieu:
Don't forget to "fixup" your commits ;)

mickaelandrieu:
no, only for review. Then it's better to have

eric-chau:
Fixup done, thanks for the review =)

Issue #165, opened 2015-01-06
[BBApplication] getClassContentDir method return the wrong vendor directory


eric-chau:
``BBApplication::getClassContentDir()`` use wrong base directory. Indeed, it use ``BBApplication::getBaseDir()`` instead of ``BBApplication::getBBDir()`` so it returns the wrong ``backbee/backbee`` class content directory

PR #166, opened 2015-01-06
[RFR] #165 + updated BBApplication::getClassContentDir

eric-chau:
Updated BBApplication::getClassContentDir() to use getBBDir method instead of getBaseDir to build path of core classcontent directory. More details: #165


eric-chau:
rebase with master done.

Issue #144, opened 2014-12-24
Bad status in authentication fail


fkroockmann:
In authentication fail, the status of response is 404 instead of 401.

Service: rest/1/security/authentication

PR #167, opened 2015-01-06
#144 Change the status code for an unknown user to 401 instead of 404

fkroockmann:
Change the status code for an unknown user to 401 instead of 404


mickaelandrieu:
Thank you @fkroockmann

eric-chau:
next time you should not merge his PR cause his commit messages are not valid!

Issue #170, opened 2015-01-07
[Validator] Don't use validator if data is empty 


fkroockmann:
On validate function it's useless to use validator for example "image or email" if the data is empty cause a property set_empty exist.

PR #171, opened 2015-01-07
[RFR] [Validator] #170 Don't use validator if data is empty

fkroockmann:
On validate function it's useless to use validator for example "image or email" if the data is empty cause a property set_empty exist.

### DIFF
--- a/Validator/ArrayValidator.php
+++ b/Validator/ArrayValidator.php
@@ -57,7 +57,7 @@ class ArrayValidator extends AValidator
                 }
 
                 if (true === $do_treatment) {
-                    if (true === isset($cConfig[self::CONFIG_PARAMETER_VALIDATOR])) {
+                    if (true === isset($cConfig[self::CONFIG_PARAMETER_VALIDATOR]) && false === empty($data)) {
                         foreach ($cConfig[self::CONFIG_PARAMETER_VALIDATOR] as $validator => $validator_conf) {
                             $this->doGeneralValidator($data, $key, $validator, $validator_conf, $errors);
                         }
--- a/Validator/ArrayValidator.php
+++ b/Validator/ArrayValidator.php
@@ -50,11 +50,13 @@ class ArrayValidator extends AValidator
         foreach ($datas as $key => $data) {
             if (null !== $cConfig = $this->getData($key, $form_config)) {
                 
-                $set_empty = isset($cConfig[self::CONFIG_PARAMETER_SET_EMPTY]) && true === $cConfig[self::CONFIG_PARAMETER_SET_EMPTY];
+                if ($set_empty = isset($cConfig[self::CONFIG_PARAMETER_SET_EMPTY])) {
+                    $set_empty =  true === $cConfig[self::CONFIG_PARAMETER_SET_EMPTY];
+                }
                 
                 $do_treatment = true;
                 if (isset($cConfig[self::CONFIG_PARAMETER_MANDATORY])) {
-                    if (false === $cConfig[self::CONFIG_PARAMETER_MANDATORY] && (true === empty($data) && false === $set_empty)) {
+                    if (false === $cConfig[self::CONFIG_PARAMETER_MANDATORY] && empty($data && !$set_empty)) {
                         $do_treatment = false;
                     }
                 }
--- a/Validator/ArrayValidator.php
+++ b/Validator/ArrayValidator.php
@@ -56,7 +56,7 @@ class ArrayValidator extends AValidator
                 
                 $do_treatment = true;
                 if (isset($cConfig[self::CONFIG_PARAMETER_MANDATORY])) {
-                    if (false === $cConfig[self::CONFIG_PARAMETER_MANDATORY] && empty($data && !$set_empty)) {
+                    if (false === $cConfig[self::CONFIG_PARAMETER_MANDATORY] && empty($data) && !$set_empty) {
                         $do_treatment = false;
                     }
                 }
--- a/Validator/ArrayValidator.php
+++ b/Validator/ArrayValidator.php
@@ -49,26 +49,36 @@ class ArrayValidator extends AValidator
     {
         foreach ($datas as $key => $data) {
             if (null !== $cConfig = $this->getData($key, $form_config)) {
+                
+                $set_empty = isset($cConfig[self::CONFIG_PARAMETER_SET_EMPTY]) && true === $cConfig[self::CONFIG_PARAMETER_SET_EMPTY];
+                
                 $do_treatment = true;
-                if (true === isset($cConfig[self::CONFIG_PARAMETER_MANDATORY]) &&
-                    false === $cConfig[self::CONFIG_PARAMETER_MANDATORY] &&
-                    true === empty($data)) {
-                    $do_treatment = false;
+                if (isset($cConfig[self::CONFIG_PARAMETER_MANDATORY])) {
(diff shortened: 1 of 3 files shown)


mickaelandrieu:
Hi @fkroockmann ,is the new behavior to be effective on others fields too?

You need to check if you try to validate an image or an email.
Also, your test suite is broken.

Start by write a test for this specific cases (image/empty fields empty), then update your feature until all tests pass.

Let us know if you need help :+1:

Validator/EntityValidator.php
```diff
@@ -1,24 +1,22 @@
 <?php
 
 /*
- * Copyright (c) 2011-2015 Lp digital system
+ * Copyright (c) 2011-2013 Lp digital system
```
eric-chau: Care, you have to rollback on it (2015, not 2013)

Validator/EntityValidator.php
```diff
@@ -1,24 +1,22 @@
...
  *
- * This file is part of BackBee.
+ * This file is part of BackBuilder5.
```
eric-chau: Same here, BackBee is right, not BackBuilder5.

Validator/EntityValidator.php
```diff
@@ -28,8 +26,8 @@
 /**
  * Entity's validator
  *
- * @category    BackBee
- * @package     BackBee\Validator
+ * @category    BackBuilder
+ * @package     BackBuilder\Validator
```
eric-chau: Here too, BackBee and not BackBuilder.

Validator/ArrayValidator.php
```diff
@@ -49,26 +49,36 @@ public function validate($array, array $datas = array(), array &$errors = array(
     {
         foreach ($datas as $key => $data) {
             if (null !== $cConfig = $this->getData($key, $form_config)) {
+                
+                $set_empty = isset($cConfig[self::CONFIG_PARAMETER_SET_EMPTY]) && true === $cConfig[self::CONFIG_PARAMETER_SET_EMPTY];
```
eric-chau: Can you limit every line to 120 characters please? Maybe something like that:

```php
    if ($setEmpty = isset($cConfig[self::CONFIG_PARAMETER_SET_EMPTY])) {
        $setEmpty =  true === $cConfig[self::CONFIG_PARAMETER_SET_EMPTY];
    }
```

Validator/ArrayValidator.php
```diff
@@ -49,26 +49,36 @@ public function validate($array, array $datas = array(), array &$errors = array(
...
+                
                 $do_treatment = true;
-                if (true === isset($cConfig[self::CONFIG_PARAMETER_MANDATORY]) &&
-                    false === $cConfig[self::CONFIG_PARAMETER_MANDATORY] &&
-                    true === empty($data)) {
-                    $do_treatment = false;
+                if (isset($cConfig[self::CONFIG_PARAMETER_MANDATORY])) {
+                    if (false === $cConfig[self::CONFIG_PARAMETER_MANDATORY] && (true === empty($data) && false === $set_empty)) {
```
eric-chau: Same here, you can reduce the number of character i think.

eric-chau:
There are some typos or coding style issues. I'm not able to judge Validator component, tests are green so I guess is good. :+1:

eric-chau:
ping @fkroockmann, tests did not pass.

eric-chau:
Thanks, :+1:

Validator/ArrayValidator.php
```diff
@@ -49,26 +49,38 @@ public function validate($array, array $datas = array(), array &$errors = array(
...
                     if (true === isset($cConfig[self::CONFIG_PARAMETER_VALIDATOR])) {
-                        foreach ($cConfig[self::CONFIG_PARAMETER_VALIDATOR] as $validator => $validator_conf) {
-                            $this->doGeneralValidator($data, $key, $validator, $validator_conf, $errors);
+                        
+                        $do_treatment = true;
+                        if (true === empty($data) && true === $set_empty) {
+                            $do_treatment = false;
+                        }
+                        
+                        if (true === $do_treatment) {
```
mickaelandrieu: snake case ? o_O gosh !

Validator/ArrayValidator.php
```diff
@@ -49,26 +49,38 @@ public function validate($array, array $datas = array(), array &$errors = array(
...
+                            }
                         }
                     }
 
                     $do_set = true;
-                    if (true === isset($cConfig[self::CONFIG_PARAMETER_SET_EMPTY])) {
-                        if (false === $cConfig[self::CONFIG_PARAMETER_SET_EMPTY] && true === empty($data)) {
-                            $do_set = false;
-                        }
+                    if (false === $set_empty && true === empty($data)) {
```
mickaelandrieu: empty() always return a boolean.

Why do I bother you with that ?

When you call a strict egality check (value and type), PHP will try to cast the right part of check with the specified type of the left part of our test.

Here, you have true which is a (bool) type so *before* check for value PHP cast empty($data) to a boolean **but** the result of ``empty()`` is always a [boolean](http://php.net/manual/fr/function.empty.php).

That's why in case of internal functions which always return boolean, you should use ``==`` and ``===`` when you need to ckeck for type of output ;)

Let me know if I'm not clear :cat2:

Validator/Tests/EntityValidatorTest.php
```diff
@@ -26,6 +26,7 @@
 use BackBee\Tests\Mock\MockBBApplication;
 use BackBee\Validator\Tests\Mock\MockEntity;
 use BackBee\Validator\Tests\Mock\MockEntity2;
+use BackBee\Validator\EntityValidator;
 
 /**
```
mickaelandrieu: Your behavior change but not your tests, in my opinion you need to write a new test that demonstrate the case you want to fix.

It's easy, mock the Validator and pass an empty value: no methods should **not** be called => ``doUniqueValidator`` , ``doPasswordValidator``, ``doGeneralValidator``.

Or, you can improve your validator this way:
*  when you call ``validate()`` init it with a ``isValidable()`` public new method when you check if you have to validate or not the input.
* then, in your tests you mock validator class and just check is isValidable() return true or false (false in this specific case)
* gosh ! your code is unit covered, I'm happy I can pay you a beer :)

What do you think ?

mickaelandrieu: ping @fkroockmann

fkroockmann: I will do this tomorrow

mickaelandrieu: It's only a suggest, you don't have to do all I suggest :lollipop:,  thank you to take a look.

eric-chau:
ping @fkroockmann, is this PR mergeable?

fkroockmann:
Not enough :/

eric-chau:
Well, it would be perfect if we can close this issue definitively.

mickaelandrieu:
hi @fkroockmann, can you squash your commits ?

eric-chau:
Don't worry @mickaelandrieu I'm on it.

eric-chau:
This PR improvements are backported into [eric-chau/Validator](https://github.com/eric-chau/Validator).

mickaelandrieu:
@eric-chau I agree to move eric-chau/Validator to backbee/Validator "if" Validator is already a dependency of BackBee :)

Do we need this dependency ?

eric-chau:
@mickaelandrieu, it's one of my own repositories at moment cause ``backbee/backbee`` does not depend on it. This was added to ``backbee/backbee`` for client project only, that's why I moved it outside of ``backbee/backbee``.

Issue #172, opened 2015-01-07
[Config] change security.yml dependency to bb5_secret_key parameter


eric-chau:
``Config/security.yml`` has dependency to ``bb5_secret_key`` from BackBee service container. We should change it to ``secret_key`` instead of ``bb5_secret_key``.

mickaelandrieu:
this issue concerns 0.1x.* branch or master branch ?

eric-chau:
``master branch`` mainly, cause we have to remove every reference to bb5 (BackBuilder5) which is the wrong project name.

PR #173, opened 2015-01-08
[RFR] [Config] #172 + updated security.yml to request 'secret_key' instead of 'bb5_secret_key'

eric-chau:
We want to remove every reference to this project old name (BackBuilder, BackBuilder5, bb5).


mickaelandrieu:
:+1:

Issue #178, opened 2015-01-13
[Bundle] fix BundleLoader::loadAndGetBundleConfigByBaseDirectory


eric-chau:
The way ``::loadAndGetBundleConfigByBaseDirectory`` compute back service id is wrong, this must be fixed.

PR #180, opened 2015-01-14
[RFR] [Bundle] #178 AbstractBaseBundle, BundleInterface and BundleLoader update to fix an issue introduced by moving core into vendor's folder

eric-chau:
Previously, we guessed bundle's service identifier from its base directory. This way works well until we move backbee into  ``vendor/``. This implies that we have no control about service identifier. To ease access to these request, we calculate bundle's service identifier with the identifier provided in application bundle configuration section.

### DIFF
--- a/Bundle/AbstractBaseBundle.php
+++ b/Bundle/AbstractBaseBundle.php
@@ -92,17 +92,8 @@ abstract class AbstractBaseBundle implements BundleInterface
         $this->application = $application;
         $bundleLoader = $this->application->getContainer()->get('bundle.loader');
 
-        if (null === $id) {
-            $this->id = basename($this->baseDir);
-        } else {
-            $this->id = $id;
-        }
-
-        if (null === $baseDir) {
-            $this->baseDir = $bundleLoader->buildBundleBaseDirectoryFromClassname(get_class($this));
-        } else {
-            $this->baseDir = $baseDir;
-        }
+        $this->baseDir = $baseDir ?: $bundleLoader->buildBundleBaseDirectoryFromClassname(get_class($this));
+        $this->id = $id ?: basename($this->baseDir);
 
         $bundleLoader->loadConfigDefinition($this->getConfigServiceId(), $this->baseDir);
 
@@ -387,17 +378,17 @@ abstract class AbstractBaseBundle implements BundleInterface
     /**
      * Returns the associated callback if "controller name/action name" couple is valid
      *
-     * @param string $controller_name the controller name (ex.: BackBee\FrontController\FrontController => front)
-     * @param string $action_name     the action name (ex.:)
+     * @param string $controllerName the controller name (ex.: BackBee\FrontController\FrontController => front)
+     * @param string $actionName     the action name (ex.: FrontController::defaultAction => default)
      *
      * @return callable|null the callback if there is one associated to "controller name/action name" couple, else null
      */
-    public function getExposedActionCallback($controller_name, $action_name)
+    public function getExposedActionCallback($controllerName, $actionName)
     {
-        $unique_name = $controller_name.'_'.$action_name;
+        $uniqueName = $controllerName.'_'.$actionName;
 
-        return array_key_exists($unique_name, $this->exposedActionsCallbacks)
-            ? $this->exposedActionsCallbacks[$unique_name]
+        return array_key_exists($uniqueName, $this->exposedActionsCallbacks)
+            ? $this->exposedActionsCallbacks[$uniqueName]
             : null
         ;
     }
@@ -409,14 +400,14 @@ abstract class AbstractBaseBundle implements BundleInterface
     {
         if (true === $this->isEnabled()) {
             $container = $this->getApplication()->getContainer();
-            foreach ((array) $this->getProperty('exposed_actions') as $controller_id => $actions) {
-                if (false === $container->has($controller_id)) {
+            foreach ((array) $this->getProperty('exposed_actions') as $controllerId => $actions) {
+                if (false === $container->has($controllerId)) {
                     throw new \InvalidArgumentException(
-                        "Exposed controller with id `$controller_id` not found for ".$this->getId()
+                        "Exposed controller with id `$controllerId` not found for ".$this->getId()
                     );
                 }
 
-                $controller = $container->get($controller_id);
+                $controller = $container->get($controllerId);
                 $this->formatAndInjectExposedAction($controller, $actions);
             }
         }
@@ -430,24 +421,24 @@ abstract class AbstractBaseBundle implements BundleInterface
      */
     private function formatAndInjectExposedAction($controller, $actions)
     {
-        $controller_id = explode('\\', get_class($controller));
-        $controller_id = str_replace('controller', '', strtolower(array_pop(($controller_id))));
-        $this->exposedActions[$controller_id] = array('actions' => array());
+        $controllerId = explode('\\', get_class($controller));
+        $controllerId = str_replace('controller', '', strtolower(array_pop(($controllerId))));
+        $this->exposedActions[$controllerId] = array('actions' => array());
 
         if ($controller instanceof BundleExposedControllerInterface) {
-            $this->exposedActions[$controller_id]['label'] = $controller->getLabel();
-            $this->exposedActions[$controller_id]['description'] = $controller->getDescription();
+            $this->exposedActions[$controllerId]['label'] = $controller->getLabel();
+            $this->exposedActions[$controllerId]['description'] = $controller->getDescription();
             array_unshift($actions, 'indexAction');
             $actions = array_unique($actions);
         }
 
         foreach ($actions as $action) {
             if (method_exists($controller, $action)) {
-                $action_id = str_replace('action', '', strtolower($action));
-                $unique_name = $controller_id.'_'.$action_id;
+                $actionId = str_replace('action', '', strtolower($action));
+                $uniqueName = $controllerId.'_'.$actionId;
 
-                $this->exposedActionsCallbacks[$unique_name] = array($controller, $action);
-                $this->exposedActions[$controller_id]['actions'][] = $action_id;
+                $this->exposedActionsCallbacks[$uniqueName] = array($controller, $action);
+                $this->exposedActions[$controllerId]['actions'][] = $actionId;
             }
         }
     }
--- a/Bundle/AbstractBaseBundle.php
+++ b/Bundle/AbstractBaseBundle.php
@@ -47,18 +47,11 @@ abstract class AbstractBaseBundle implements BundleInterface
     private $application;
 
     /**
-     * Bundle loader of current application
-     *
-     * @var BackBee\Bundle\BundleLoader
-     */
-    private $bundle_loader;
-
-    /**
      * Bundle base directory
      *
      * @var string
      */
-    private $base_directory;
+    private $baseDir;
 
     /**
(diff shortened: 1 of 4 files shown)


eric-chau:
ping @backbee/core-team, I need to merge it very soonly. Thanks.

Bundle/AbstractBaseBundle.php
```diff
@@ -94,27 +80,35 @@
...
-    private $exposed_actions_callbacks;
+    private $exposedActionsCallbacks;
 
     /**
      * AbstractBaseBundle's constructor
      *
      * @param ApplicationInterface $application application to link current bundle with
      */
-    public function __construct(ApplicationInterface $application)
+    public function __construct(ApplicationInterface $application, $id = null, $baseDir = null)
```
mickaelandrieu: This part of BackBee is not covered by tests I guess.

Bundle/AbstractBaseBundle.php
```diff
@@ -94,27 +80,35 @@
...
     {
         $this->application = $application;
-        $this->bundle_loader = $this->application->getContainer()->get('bundle.loader');
+        $bundleLoader = $this->application->getContainer()->get('bundle.loader');
 
-        $this->base_directory = $this->bundle_loader->buildBundleBaseDirectoryFromClassname(get_class($this));
-        $this->id = basename($this->base_directory);
+        if (null === $id) {
```
mickaelandrieu:
```php
$id = (null === $id) ? basename($this->baseDir) : $id;
```

Bundle/AbstractBaseBundle.php
```diff
@@ -94,27 +80,35 @@
...
+            $this->id = basename($this->baseDir);
+        } else {
+            $this->id = $id;
+        }
 
-        $this->bundle_loader->loadConfigDefinition($this->getConfigServiceId(), $this->base_directory);
+        if (null === $baseDir) {
```
mickaelandrieu:
```php
$this->baseDir = (null === $baseDir) ? $bundleLoader->buildBundleBaseDirectoryFromClassname(get_class($this)) : $baseDir;
```

Bundle/AbstractBaseBundle.php
```diff
@@ -171,12 +165,11 @@ public function getConfig()
      */
     public function getConfigServiceId()
```
mickaelandrieu: improvement idea because it's in public API: can allow to pass an ``$serviceId``

Bundle/AbstractBaseBundle.php
eric-chau: yeah don't know why I didn't do it like that, apologize. We can even go this way:
```php
$this->id = $id ?: basename($this->baseDir);
```

Bundle/AbstractBaseBundle.php
```diff
@@ -403,8 +396,8 @@ public function getExposedActionCallback($controller_name, $action_name)
     {
         $unique_name = $controller_name.'_'.$action_name;
 
-        return array_key_exists($unique_name, $this->exposed_actions_callbacks)
-            ? $this->exposed_actions_callbacks[$unique_name]
+        return array_key_exists($unique_name, $this->exposedActionsCallbacks)
```
mickaelandrieu: $uniqueName

Bundle/AbstractBaseBundle.php
eric-chau: Not sure I get what you want to mean. You want me to add an argument called ``$serviceId``? This should not be here cause getConfigServiceId() must returns same value, it's not a util method but an accessor of a bundle object.

Bundle/BundleLoader.php
```diff
@@ -53,119 +52,114 @@ class BundleLoader
...
+        if (false === array_key_exists($classname, $this->reflectionClasses)) {
+            $this->reflectionClasses[$classname] = new \ReflectionClass($classname);
         }
 
-        $base_directory = dirname($this->reflection_classes[$classname]->getFileName());
+        $baseDir = dirname($this->reflectionClasses[$classname]->getFileName());
 
-        if (false === is_dir($base_directory)) {
+        if (false === is_dir($baseDir)) {
             throw new \Exception();
```
mickaelandrieu: Maybe a message can be useful here ?

Bundle/BundleLoader.php
```diff
@@ -173,11 +167,15 @@ private function generateBundleServiceId($id)
...
+     * @param  string $classname The bundle entry point classname
+     * @param  string $bundleId  The bundle id/name
+     * @param  string $baseDir   The bundle base directory
+     * @return \Symfony\Component\DependencyInjection\Definition
+     * @throws InvalidArgumentException if provided classname does not implements BackBee\Bundle\BundleInterface
      */
-    private function buildBundleDefinition($classname, $base_directory)
+    private function buildBundleDefinition($classname, $bundleId, $baseDir)
     {
         if (false === in_array('BackBee\Bundle\BundleInterface', class_implements($classname))) {
```
mickaelandrieu: use ``is_subclass_of`` instead, more performant => http://bg2.php.net/manual/en/function.is-subclass-of.php

Bundle/BundleLoader.php
```diff
@@ -193,85 +191,63 @@ private function buildBundleDefinition($classname, $base_directory)
...
     private function loadFullBundles()
     {
-        foreach ($this->bundles_base_directory as $base_directory) {
-            $config = $this->loadAndGetBundleConfigByBaseDirectory($base_directory);
-            $bundle_config = $config->getSection('bundle');
-            if (isset($bundle_config['enable']) && false === (boolean) $bundle_config['enable']) {
+        foreach ($this->bundlesBaseDir as $serviceId => $baseDir) {
+            $config = $this->loadAndGetBundleConfigByBaseDir($serviceId, $baseDir);
+            $bundleConfig = $config->getSection('bundle');
+            if (isset($bundleConfig['enable']) && false === (boolean) $bundleConfig['enable']) {
```
mickaelandrieu: you don't need ``===`` if you cast into boolean :smile_cat:

Bundle/BundleLoader.php
```diff
@@ -193,85 +191,63 @@ private function buildBundleDefinition($classname, $base_directory)
...
-            );
+            $this->loadServices($config, $this->getCallbackFromRecipes(self::SERVICE_RECIPE_KEY, $recipes));
+            $this->loadEvents($config, $this->getCallbackFromRecipes(self::EVENT_RECIPE_KEY, $recipes));
+            $this->loadRoutes($config, $this->getCallbackFromRecipes(self::ROUTE_RECIPE_KEY, $recipes));
+            $this->addClassContentDir($config, $this->getCallbackFromRecipes(self::CLASSCONTENT_RECIPE_KEY, $recipes));
+            $this->addTemplatesDir($config, $this->getCallbackFromRecipes(self::TEMPLATE_RECIPE_KEY, $recipes));
+            $this->addHelpersDir($config, $this->getCallbackFromRecipes(self::HELPER_RECIPE_KEY, $recipes));
+            $this->addResourcesDir($config, $this->getCallbackFromRecipes(self::RESOURCE_RECIPE_KEY, $recipes));
+            $this->addNamespaces($config, $this->getCallbackFromRecipes(self::NAMESPACE_RECIPE_KEY, $recipes));
+            $this->runRecipe($config, $this->getCallbackFromRecipes(self::CUSTOM_RECIPE_KEY, $recipes));
```
mickaelandrieu: should be ``runRecipes()`` right ?

don't forget to rename it

Bundle/BundleLoader.php
```diff
@@ -193,85 +191,63 @@ private function buildBundleDefinition($classname, $base_directory)
...
-        $config_id = str_replace('%bundle_id%', basename($base_directory), BundleInterface::CONFIG_SERVICE_ID_PATTERN);
-        $this->loadConfigDefinition($config_id, $base_directory);
-        $bundle_config = $this->container->get($config_id)->getBundleConfig();
-        if (true === isset($bundle_config['config_per_site']) && true === $bundle_config['config_per_site']) {
-            $definition = $this->container->getDefinition($config_id);
+        $configId = str_replace('%bundle_service_id%', $serviceId, BundleInterface::CONFIG_SERVICE_ID_PATTERN);
+
+        $this->loadConfigDefinition($configId, $baseDir);
+        $bundleConfig = $this->container->get($configId)->getBundleConfig();
+        if (true === isset($bundleConfig['config_per_site']) && true === $bundleConfig['config_per_site']) {
```
mickaelandrieu: you don't need ``true === isset()`` but ``if(isset() && true === ...)`` because [isset](http://php.net/manual/fr/function.isset.php) always return a boolean value.

Bundle/BundleLoader.php
```diff
@@ -291,49 +267,50 @@ private function buildConfigDefinition($base_directory)
...
      *
-     * @return [type] [description]
+     * @return string
      */
-    private function getConfigDirectoryByBaseDirectory($base_directory)
+    private function getConfigDirByBundleBaseDir($baseDir)
     {
-        $directory = $base_directory.DIRECTORY_SEPARATOR.BundleInterface::CONFIG_DIRECTORY_NAME;
+        $directory = $baseDir.DIRECTORY_SEPARATOR.BundleInterface::CONFIG_DIRECTORY_NAME;
         if (false === is_dir($directory)) {
```
mickaelandrieu: Same here, I may forget some ``is_dir`` checks before.

Bundle/BundleLoader.php
```diff
@@ -291,49 +267,50 @@ private function buildConfigDefinition($base_directory)
...
+     * @return array|null
      */
     private function getLoaderRecipesByConfig(Config $config)
     {
         $recipes = null;
-        $bundle_config = $config->getBundleConfig();
-        if (null !== $bundle_config && true === array_key_exists('bundle_loader_recipes', $bundle_config)) {
-            $recipes = $bundle_config['bundle_loader_recipes'];
+        $bundleConfig = $config->getBundleConfig();
+        if (null !== $bundleConfig && array_key_exists('bundle_loader_recipes', $bundleConfig)) {
```
mickaelandrieu: :+1: perfect !

Bundle/AbstractBaseBundle.php
mickaelandrieu: Can you change method to ``protected`` ?

Bundle/AbstractBaseBundle.php
eric-chau: You're right. Writing tests for Bundle component will require alot of time, I will do it when i got it. Ofc if somebody want to write it he is welcome =)

Bundle/BundleLoader.php
mickaelandrieu: did you see this comment ? It's (very) important to give explicit name to functions.

eric-chau: Yeah I saw it but everything is fine here. Instead of calling ``runRecipe()`` i could create a method (something like ``runCustomRecipe()``) but it would just be an alias to ``runRecipe()``, that why I decide to do it like this. If you look closer you will notice that every methods like ``loadEvents()`` or ``addNamespaces()`` called internally ``runRecipe()``. But these has specific method cause they have a default behavior if developer didn't want to perform his custom recipe. Let me know if it's not clear.

mickaelandrieu: Did you run one recipe or multiple recipes ? I misunderstand why we have plural "method name" on all previous functions but not in the last.

Issue #179, opened 2015-01-13
Remove old backbee toolbar source files


eric-chau:
It includes Php, Javascript and CSS files.

New BackBee\ToolbarBundle will embed its own source files.

mickaelandrieu:
What is the version(s) of BackBee concerned by this issue ?

eric-chau:
BackBee master (min version 0.12)

mickaelandrieu:
@eric-chau is this issue solved ?

PR #181, opened 2015-01-14
[RFR] #179 removed source files (php, js, css) of old backbee toolbar

eric-chau:
Removed:
- ``Resources/js/`` folder which contained toolbar Javascript files and aloha's files
- ``Resources/css/``folder cause backbee core should not contains css like bootstrap 3 or font awesome.
- ``Services/`` RPC Json Server and its services local

Remaining deletions:
- ``Resources/img``, maybe not the whole folder but most of image in this folder must be deleted.


mickaelandrieu:
:+1:

Issue #196, opened 2015-01-20
Update project dependencies


eric-chau:
Atleast Symfony and Twig are updated (respectively to ``2.3.24`` and ``1.17.0``).

PR #198, opened 2015-01-20
[RFR] #196 + updated composer.json and project's dependencies

eric-chau:
Some updates are very effective like ``twig`` which improve the cache management or ``symfony`` which brings improvements about security.

Note that ``doctrine/dbal`` breaks compatibilty with backbee since version >= 2.5. I didn't test with ``doctrine/orm`` version 2.5 or later yet cause it still in alpha release.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,17 +16,17 @@
     ],
     "require": {
         "php": ">=5.4.0",
-        "symfony/symfony": "2.3.23",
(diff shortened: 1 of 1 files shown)


mickaelandrieu:
:+1: I will take a look to doctrine >= 2.5 because I'm highly interested by the new cache they provide :)

PR #200, opened 2015-01-20
[RFR] update dependencies and annotations

mickaelandrieu:
* Removed symfony/symfony from project
* Added only components used
* Udpated to 2.5.x (except validator component because of BC of API)
* Updated path to search Validator annotations on our autoloader
* Pray a little bit

In my tests, testsuite is 100% faster (24s from 57s)


mickaelandrieu:
ping @crouillon , we should not wait too much to merge this one in order to have time to fix issues if required before the 1.0 release.

Issue #177, opened 2015-01-13
[Rest] add management of classcontent update and its revisions


eric-chau:
Create ``BackBee\Rest\Controller\ClassContentController::putAction`` and ``RevisionController``?

Waiting for PR #109 to be merged into master branch to start.

Save of classcontent includes elements and parameters so this issue will also resolve #33: Store only parameters overiding in content table.

eric-chau:
This is an example of a concise representation of an article:
```json
{
    "uid": "d6967c034dbf3a26c9eedfd394555364",
    "type": "article",
    "parameters": {
        "excludefromautobloc": {
            "array": {
                "rendertype": "checkbox",
                "label": "Exclude from autobloc",
                "checked": false
            }
        }
    },
    "elements": {
        "title": {
            "uid": "b0f1b56ef7a5edf55039091a171ed993",
            "type": "Element/text"
        },
        "title_alt": {
            "uid": "9e80aa7bde1dd8ad1170f262317a7574",
            "type": "Element/text"
        },
        "permid": "",
        "abstract": {
            "uid": "b4b4c8e8639b1a5ea15da0824173d90e",
            "type": "Element/text"
        },
        "body": {
            "uid": "560f375d627b574409657bdaac052cd8",
            "type": "article/body"
        },
        "image": {
            "uid": "44a19a363763940d3481f457ab3a135d",
            "type": "Media/image"
        },
        "image_alt": {
            "uid": "179a4b0f752316dfc61e0b84b290f7aa",
            "type": "Media/image"
        },
        "keywords": {
            "uid": "d833bd9761a0ec19057a601bd556d330",
            "type": "Element/keyword"
        },
        "related": {
            "uid": "e2a5331a57d356a1acd452414112ff9d",
            "type": "article/related"
        },
        "signature": ""
    },
    "image": "http://backbee.dev:8000/resources/img/contents/default_thumbnail.png"
}
```
If you want to change the __title_alt__ element to be equals to __title__ text element, you have to perform this request ``PUT /rest/1/classcontent/article/d6967c034dbf3a26c9eedfd394555364`` with these data:
```json
{
    "elements": {
        "title_alt": {
            "uid": "b0f1b56ef7a5edf55039091a171ed993",
            "type": "Element/text"
        }
    }
}
```

This is a concise representation of the article's title element:
```json
{
    "uid": "b0f1b56ef7a5edf55039091a171ed993",
    "type": "Element/text",
    "parameters": [],
    "elements": {
[...]

PR #211, opened 2015-01-23
[RFR] [Rest] #177 add management of classcontent update and its revisions

eric-chau:
More details at #177.


ClassContent/AContent.php
```diff
@@ -162,14 +188,18 @@ public function __construct($uid = null, $options = null)
...
+     * @param  string $var The name of the element
+     * @return mixed The value
+     * @throws UnknownPropertyException Occurs when $var does not match an element
      */
     public function __get($var)
     {
         if ($this->_getContentInstance() instanceof ContentSet) {
-            throw new Exception\UnknownPropertyException(sprintf('Unknown property %s in %s.', $var, ClassUtils::getRealClass($this->_getContentInstance())));
+            throw new UnknownPropertyException(sprintf(
+                'Unknown property %s in %s.',
```
mickaelandrieu: You have an additional information to return: the $var is a ContentSet (with probably an uuid).

Think: "How can I do to help myself to debug if my application is broken ?"

ClassContent/AContent.php
```diff
@@ -194,11 +228,22 @@ public function __set($var, $value)
         $val = array();
 
         foreach ($values as $value) {
-            if ((isset($this->_maxentry[$var]) && 0 < $this->_maxentry[$var] && $this->_maxentry[$var] == count($val)) || (isset($this->_minentry[$var]) && count($val) < $this->_minentry[$var] && $this->_maxentry[$var] == count($val))) {
+            if (
+                (
+                    isset($this->_maxentry[$var])
```
mickaelandrieu: ``$this->maxentry`` if you can without too much work :)

ClassContent/AContent.php
```diff
@@ -571,8 +631,44 @@ public function acceptSubcontent($var)
...
+
+    /**
+     * Checks for an accepted type
+     * @param  mixed   $value the value from which the type will be checked
+     * @param  string  $var   the element to be checks
+     * @return Boolean
+     */
+    public function isAccepted($value, $var = null)
+    {
+        if ($this->_getContentInstance() instanceof ContentSet) {
```
mickaelandrieu: are you sure to receive an Object at this point ?

ClassContent/AContent.php
```diff
@@ -571,8 +631,44 @@ public function acceptSubcontent($var)
...
+            if (false === ($value instanceof AClassContent)) {
```
mickaelandrieu: ``instanceof`` always return a boolean, or Fatal exception if input is not an object.

ClassContent/AContent.php
eric-chau: ``getContentInstance()`` will always return an object yeah.

ClassContent/AContent.php
```diff
@@ -571,8 +631,44 @@ public function acceptSubcontent($var)
...
+                return false;
+            }
+
+            $accept_array = $this->_accept;
+        } else {
+            if (null === $var) {
+                return false;
+            }
+
+            if (false === array_key_exists($var, $this->_accept)) {
```
mickaelandrieu: same here :)

ClassContent/ClassContentManager.php
```diff
@@ -0,0 +1,321 @@
+<?php
+
+/*
+ * Copyright (c) 2011-2015 Lp digital system
+ *
+ * This file is part of BackBee.
+ *
+ * BackBee5 is free software: you can redistribute it and/or modify
```
mickaelandrieu: I need to make a PR for this one, we still have "BackBee5" occurrences :(

eric-chau:
@mickaelandrieu, thanks for your code review but don't spam me about ``_`` appended to method or class' attributes cause I'm not aiming to do these changes at moment. I want to build up every service needed by @fkroockmann before focusing on this task. Thanks =)

eric-chau:
But be sure I will update all these things when i'm done and so we can discuss about it ^^

ClassContent/AContent.php
```diff
@@ -735,7 +798,7 @@ public function getIdentifier()
      */
     public function getType()
     {
-        return ClassUtils::getRealClass($this);
+        return str_replace([self::CLASSCONTENT_BASE_NAMESPACE, '\\'], ['', '/'], ClassUtils::getRealClass($this));
     }
```
crouillon: Hum, change \ to / seems to be hasardous here

ClassContent/ClassContentManager.php
```diff
@@ -0,0 +1,321 @@
...
+
+        return array_merge($this->getAllElementClassContentClassnames(), $classnames);
+    }
+
+    /**
+     * Returns classnames of all classcontents element
+     *
+     * @return array Contains every BackBee's element classcontent classnames
+     */
+    public function getAllElementClassContentClassnames()
```
mickaelandrieu: Nice additions, but where are the tests ? :'(

ClassContent/AContent.php
eric-chau: I don't get what you want to mean @crouillon, can you give me more details please?

ClassContent/ClassContentManager.php
```diff
@@ -0,0 +1,321 @@
...
+    {
+        $directory = $this->application->getBBDir().DIRECTORY_SEPARATOR.'ClassContent';
```
mickaelandrieu: you should add a $path in function declaration to allow user to set another path.

pros: esay to test, easy to re-use
cans: you need a check for the path inside function, maybe useless ? => function should be protected or private.

ClassContent/ClassContentManager.php
eric-chau: Like I tagged, this work still in progress :D

ClassContent/ClassContentManager.php
eric-chau: I get what you want to mean about adding ``$path`` parameter to this method but i don't really get about the mentionned convenient. Can you tell me more about it please?

Rest/Controller/ClassContentController.php
```diff
@@ -174,28 +171,20 @@ public function getAction($type, $uid, Request $request)
...
-        $draft = $em->getRepository('BackBee\ClassContent\Revision')->getDraft(
-            $content,
-            $this->getApplication()->getBBUserToken(),
-            true
-        );
-
-        $content->setDraft($draft);
-        $em->flush();
+        $this->getEntityManager()->persist($content);
+        $content->setDraft($this->getManager()->getRevision($content, true));
```
mickaelandrieu: What is the ``getManager()`` ? We should use a more specific name.

Rest/Controller/ClassContentController.php
```diff
@@ -205,21 +194,36 @@ public function postAction($type, Request $request)
...
     /**
+     * Updates classcontent's elements and parameters
+     *
+     * @param  string $type
+     * @param  string $uid
+     * @return Symfony\Component\HttpFoundation\JsonResponse
+     */
+    public function putAction($type, $uid, Request $request)
+    {
+        $this->granted('EDIT', $content = $this->getManager()->findOneByTypeAndUid($type, $uid, true, true));
```
mickaelandrieu: same here (to no forget)

Rest/Controller/ClassContentController.php
```diff
@@ -205,21 +194,36 @@ public function postAction($type, Request $request)
...
      * @param string $type type of the class content (ex: Element/text)
      * @param string $uid  identifier of the class content
-     *
-     * @return Symfony\Component\HttpFoundation\Response
+     * @return Symfony\Component\HttpFoundation\JsonResponse
      */
     public function deleteAction($type, $uid)
     {
-        $content = $this->getClassContentByTypeAndUid($type, $uid);
+        $content = $this->getManager()->findOneByTypeAndUid($type, $uid);
```
mickaelandrieu: this is a ``ContentManager``, right ?

Rest/Controller/ClassContentController.php
```diff
@@ -241,40 +245,32 @@ private function getCategoryManager()
...
      *
-     * @param string $type
-     *
-     * @return string classname associated to provided
-     *
-     * @throws
+     * @return BackBee\ClassContent\ClassContentManager
      */
-    private function getClassnameByType($type)
+    private function getManager()
```
mickaelandrieu: finally found: would you mind to rename it "ClassContentManager`` ? auto-complete will do the work, don't worry :)

ClassContent/ClassContentManager.php
```diff
@@ -0,0 +1,321 @@
...
+     * @param  AClassContent $content
+     * @param  integer       $format
+     * @return array
+     */
+    public function jsonEncode(AClassContent $content, $format = AClassContent::JSON_DEFAULT_FORMAT)
+    {
+        if (AClassContent::JSON_DEFINITION_FORMAT === $format) {
+            $classname = get_class($content);
+            $content = new $classname;
+        }
```
crouillon: That was the reason AclassContent::getType() was existing: with entities managed by Doctrine get_class() can returns either the real classname or the proxy classname,

Rest/Controller/ClassContentController.php
eric-chau: I consider that in a ``ClassContentController`` we can easily guess that ``::getManager`` will returns a ClassContentManager. In addition I also documented the method and its type with docblock of the class attribute. I could call it something like ``getClassContentManager`` but it's kinda too long, isn't it?

ClassContent/ClassContentManager.php
mickaelandrieu: How about the ``ClassContents`` provided by Bundles ? this function can't return them, right ?

Then, you will meet some difficulty to test it because you need to create ``/repository/ClassContent/`` folder into your mocked application.

ClassContent/ClassContentManager.php
eric-chau: Hm ok. Since ``getType`` API changed, may I change it from:

```php
$classname = get_class($content);
$content = new $classname;
```
to:
```php
$classname = ClassUtils::getRealClass(get_class($content));
$content = new $classname;
```
What do you think about it @crouillon?

Rest/Controller/ClassContentController.php
mickaelandrieu: Nope (I will remove the next comments).

Name lenght of variable is never a good argument face to the understanding of your code.
Maybe ``getContentManager`` is enough, but ``getManager()`` is not: what do we manage at this point ? is it a global manager ? a content manager ?  a manager manager ?

ClassContent/ClassContentManager.php
eric-chau: It's quite easy to do so cause current mock of ``BBApplication`` has also a mocked folder tree.

ClassContent/AContent.php
crouillon: with Bloc/article.yml
before : article->geType() returns BackBee\ClassContent\Bloc\article
after : article->geType() returns Boc/article
Being an existing public function, it is hasardous to change so much its return

Rest/Controller/ClassContentController.php
eric-chau: Then i will renamed it to ``getClassContentManager`` even it sounds logical to me.

ClassContent/ClassContentManager.php
crouillon: See my previous note, why change the getType API?

ClassContent/AContent.php
eric-chau: I added this major change into ``CHANGELOG-0.12.md``. Maybe we can just add a new method like ``AContent::getRealClassname()`` or something like that which return the real classname and not the doctrine proxy one. If we don't do so how to name the method which return the value ``Bloc/article`` as this content's type?

ClassContent/ClassContentManager.php
eric-chau: I did it cause it sounds more logical to me to do it like that. What we did with previous ``getType`` is to change behavior of Php ``get_class`` and to rename it to type which is quite strange, isn't it?

ClassContent/AContent.php
crouillon: AContent and so AClassContent implements \BackBuilder\Security\Acl\Domain\IObjectIdentifiable to allow rights definition on content, change the return for getType() is a problem i think

eric-chau: Ok I didn't noticed it. Then I don't know how to call this. ``getShortenNamespace``? ``getContentType``?

ClassContent/ClassContentManager.php
eric-chau: This was fixed by adding ``AContent::getContentType``.

fkroockmann:
When you delete a parameter, the REST call retrieve yet the parameter.

eric-chau:
ping @backbee/core-team

ClassContent/AContent.php
```diff
@@ -570,9 +594,39 @@ public function acceptSubcontent($var)
...
+
+    /**
+     * Checks for an accepted type
+     * @param  mixed   $value the value from which the type will be checked
+     * @param  string  $var   the element to be checks
+     * @return Boolean
+     */
+    public function isAccepted($value, $var = null)
+    {
+        if ($this->getContentInstance() instanceof ContentSet) {
```
mickaelandrieu: hmm, this need some comments.

eric-chau: You need comments about ``$this->getContentIntance()``? Or something else?

Issue #223, opened 2015-01-30
[ClassContent] fix IndexationRepository for HHVM and php7


ndufreche:
Before some tests, the problem with HHVM and php7 become from the method getParentContentUids in IndexationRepository.

PR #224, opened 2015-01-30
[ClassContent] fix IndexationRepository for HHVM and php7 #223

ndufreche:
Before some tests, the probleme HHVM and php7 become from the method getParentContentUids in IndexationRepository.


mickaelandrieu:
:+1:

Issue #225, opened 2015-01-30
[rest] Complete rest user controller


ndufreche:
New update :
- [ ] A User have to return the groups they are part
- [ ] Update getCollectionAction with adding filters
  - [ ] Group filter
- [ ] Develop patchAction
  - [ ] Patch user group

ndufreche:
The user controller still in development.

mickaelandrieu:
what is the status of this issue ?

eric-chau:
@ndufreche can we close this issue?

ndufreche:
Yes :)

eric-chau:
Thank you =)

PR #243, opened 2015-02-11
[rest] Complete rest user controller #225

ndufreche:
This not complete implementation of the issue #225 but is needed by backbee/BbCoreJs#17

### DIFF
--- a/Rest/Controller/UserController.php
+++ b/Rest/Controller/UserController.php
@@ -59,14 +59,13 @@ class UserController extends ARestController
      */
     public function getCollectionAction(Request $request)
     {
-        // TODO
-
-
         if (!$this->isGranted('VIEW', new ObjectIdentity('class', 'BackBee\Security\User'))) {
             throw new AccessDeniedException(sprintf('You are not authorized to view users'));
         }
 
-        return array();
+        $users = $this->getEntityManager()->getRepository('BackBee\Security\User')->findAll();
+
+        return new Response($this->formatCollection($users), 200, ['Content-Type' => 'application/json']);
     }
 
     /**
@@ -120,6 +119,52 @@ class UserController extends ARestController
         return new Response("", 204);
     }
 
+    public function checkIdentity($username, $password)
+    {
+        $created = date('Y-m-d H:i:s');
+        $token = new BBUserToken();
+        $token->setUser($request->request->get('username'));
+        $token->setCreated($created);
+        $token->setNonce(md5(uniqid('', true)));
+        $token->setDigest(md5($token->getNonce() . $created . md5($password)));
+
+        $tokenAuthenticated = $this->getApplication()->getSecurityContext()->getAuthenticationManager()
+            ->authenticate($token)
+        ;
+
+        $this->getApplication()->getSecurityContext()->setToken($tokenAuthenticated);
+    }
+
+
+    /**
+     * PATCH User
+     *
+     * @param int $id User ID
+     */
+    public function patchAction($id, Request $request)
+    {
+        if (!$this->isGranted('IS_AUTHENTICATED_FULLY')) {
+            throw new AccessDeniedException('You must be authenticated to view users');
+        }
+
+        $user = $this->getApplication()->getBBUserToken()->getUser();
+
+        if ($user->getId() !== $id) {
+            throw new AccessDeniedException('Password can only be changed by its owner.');
+        }
+
+        if ($user->getState() === User::PASSWORD_NOT_PICKED) {
+            $this->updateUserPassword($user, $request);
+        } else {
+            $oldPassword = $request->get('old_password', '');
+
+            $this->checkIdentity($user->getLogin(), $oldPassword);
+            $this->updateUserPassword($user, $request);
+        }
+
+        return new Response("", 204);
+    }
+
     /**
      * UPDATE User
      *
@@ -127,11 +172,9 @@ class UserController extends ARestController
      *  @Assert\NotBlank(message="Login is required"),
      *  @Assert\Length(min=6, minMessage="Minimum length of the login is 6 characters")
      * })
-     * @Rest\RequestParam(name = "firstname", requirements = {
-     *  @Assert\NotBlank(message="First Name is required")
-     * })
-     * @Rest\RequestParam(name = "lastname", requirements = {
-     *  @Assert\NotBlank(message="Last Name is required")
+     * @Rest\RequestParam(name = "email", requirements = {
+     *  @Assert\NotBlank(message="Email not provided"),
+     *  @Assert\Email(checkMX=true, message="Email invalid")
      * })
      *
      * @param int $id User ID
@@ -152,10 +195,21 @@ class UserController extends ARestController
             throw new AccessDeniedException(sprintf('You are not authorized to view user with id %s', $id));
         }
 
-        $this->deserializeEntity($request->request->all(), $user);
+        $user = $this->deserializeEntity($request->request->all(), $user);
+
+        if ($request->request->has('password')) {
+            $encoderFactory = $this->getContainer()->get('security.context')->getEncoderFactory();
+            $password = $request->request->get('password', '');
+
+            if ($encoderFactory && $encoder = $encoderFactory->getEncoder($user)) {
+                $password = $encoder->encodePassword($password, '');
+            }
+
+            $user->setPassword($password);
+        }
 
         $this->getEntityManager()->persist($user);
-        $this->getEntityManager()->flush();
+        $this->getEntityManager()->flush($user);
 
         return new Response("", 204);
     }
@@ -163,26 +217,17 @@ class UserController extends ARestController
     /**
      * Create User
      *
-     *
-     * @Rest\RequestParam(name = "password", requirements = {
-     *  @Assert\NotBlank(message="Password not provided"),
-     *  @Assert\Length(min=6, minMessage="Password minimum length is 6 characters")
-     * })
      * @Rest\RequestParam(name = "login", requirements = {
      *  @Assert\NotBlank(message="Login is required"),
-     *  @Assert\Length(min=6, minMessage="Minimum length of the login is 6 characters")
      * })
-     * @Rest\RequestParam(name = "firstname", requirements = {
-     *  @Assert\NotBlank(message="First Name is required")
+     * @Rest\RequestParam(name = "email", requirements = {
+     *  @Assert\NotBlank(message="Email not provided"),
+     *  @Assert\Email(checkMX=true, message="Email invalid")
      * })
-     * @Rest\RequestParam(name = "lastname", requirements = {
-     *  @Assert\NotBlank(message="Last Name is required")
-     * })
-     *
-     *
      */
     public function postAction(Request $request)
     {
+
         if (!$this->isGranted('IS_AUTHENTICATED_FULLY')) {
             throw new AccessDeniedException('You must be authenticated to view users');
         }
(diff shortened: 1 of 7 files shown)


Rest/Controller/UserController.php
```diff
@@ -120,6 +119,43 @@ public function deleteAction($id)
...
+     *
+     * @param int $id User ID
+     */
+    public function patchAction($id, Request $request)
+    {
+        if (!$this->isGranted('IS_AUTHENTICATED_FULLY')) {
+            throw new AccessDeniedException('You must be authenticated to view users');
+        }
+
+        $user = $this->getApplication()->getBBUserToken()->getUser()-getId();
```
mickaelandrieu: ``$userId `` you mean ?

Rest/Controller/UserController.php
```diff
@@ -120,6 +119,43 @@ public function deleteAction($id)
...
+
+        if ($user->getId() !== $id) {
```
mickaelandrieu: you probably return $userId on the previous line, so you can't access user id this way.

Rest/Controller/UserController.php
```diff
@@ -152,7 +186,18 @@ public function putAction($id, Request $request)
...
         }
 
-        $this->deserializeEntity($request->request->all(), $user);
+        $user = $this->deserializeEntity($request->request->all(), $user);
+
+        if ($request->request->has('password')) {
+            $encoderFactory = $this->getContainer()->get('security.context')->getEncoderFactory();
+            $password = $request->request->get('password', '');
+
+            if ($encoderFactory && $encoder = $encoderFactory->getEncoder($user)) {
```
mickaelandrieu: you probably don't need all this checks... can you explain ?

Rest/Controller/UserController.php
```diff
@@ -200,22 +236,52 @@ public function postAction(Request $request)
...
 
         $user = $this->deserializeEntity($request->request->all(), $user);
-
         // handle the password
         if ($request->request->has('password')) {
-            $encoderFactory = $this->getContainer()->get('security.context')->getEncoderFactory();
-            $password = $request->request->get('password');
+           $password = $request->request->get('password');
+        } else {
+            $password = substr(hash('sha512', rand()), 0, 6);
```
mickaelandrieu: o_O if user don't return password, you automaticaly set a generated ? imo you should return an exception here.

Rest/Controller/UserController.php
```diff
@@ -200,22 +236,52 @@ public function postAction(Request $request)
...
+        if ($encoderFactory && $encoder = $encoderFactory->getEncoder($user)) {
+            $password = $encoder->encodePassword($password, '');
         }
 
+        $user->setPassword($password);
+
         $this->getEntityManager()->persist($user);
         $this->getEntityManager()->flush();
 
         return new Response($this->formatItem($user), 200, array('Content-Type' => 'application/json'));
```
mickaelandrieu: be a brave, use ``[ ]`` notation instead :)

Rest/Controller/UserController.php
```diff
@@ -200,22 +236,52 @@ public function postAction(Request $request)
...
+     * @todo  set minimal password size in secirity config
+     * [updateUserPassword description]
+     * @param  User    $user    [description]
+     * @param  Request $request [description]
+     * @return [type]           [description]
+     */
+    private function updateUserPassword(User $user, Request $request)
+    {
+        if ($request->request->get('password') !== $request->request->get('confirm_password')) {
+            throw new AccessDeniedException(sprintf('password and confirm password are differents'));
```
mickaelandrieu: you don't need ``sprintf`` here.

Rest/Controller/UserController.php
```diff
@@ -200,22 +236,52 @@ public function postAction(Request $request)
...
+        }
+        $password = trim($request->request->get('password'));
+        if (strlen($password) < 5) {
+            throw new AccessDeniedException(sprintf('Password to short.'));
```
mickaelandrieu: and here, obvsly.

Security/User.php
```diff
@@ -40,13 +40,16 @@
  */
 class User implements UserInterface
 {
+    const PASSWORD_NOT_PICKED = 0;
+
```
mickaelandrieu: remove unused line please.

Security/User.php
```diff
@@ -232,6 +261,32 @@ public function setLogin($login)
 
     /**
      *
+     * @param  string                 $login
+     * @return \BackBee\Security\User
+     * @codeCoverageIgnore
```
mickaelandrieu: why this is ignored by code coverage ?

mickaelandrieu:
Also, you may have removed a test about private key I'm not realy sure.

Rest/Controller/UserController.php
ndufreche: I don't return an user id, Mysterious PHP i don't know why this line work :
```php
$user = $this->getApplication()->getBBUserToken()->getUser()-getId();
```

Rest/Controller/UserController.php
ndufreche: It works, xD. But why ??

Rest/Controller/UserController.php
ndufreche: Nop, i take that from @pasinter  implementation.

Rest/Controller/UserController.php
ndufreche: Yes it's a request from @crouillon

Security/User.php
ndufreche: The carriage return ?

ndufreche:
I removed the test on private key because it was impossible of this test have be worked a day.

Security/User.php
ndufreche: Because this is a copy/paste of the comment setLogin function

Rest/Controller/UserController.php
mickaelandrieu: o_O realy ? w t f ...

Rest/Controller/UserController.php
mickaelandrieu: anyway, can you remove ``-getId()`` please ? :+1:

Security/User.php
mickaelandrieu: something like 
```php
const FOO = 'foo';
const BAR = 'bar';
```
without empty line between constants. Do you agree ?

mickaelandrieu:
It does, we are all green since end of december. You can skip the test for now, but not delete it ;)

ndufreche:
All test are green from December yes. But I goes explain you why the test can't work :
In the user definition we have that :
```php
    /**
     * User's private api key
     * @var String
     * @Column(type="string", name="api_key_private", nullable=true)
     * @fixture(type="string")
     * @Serializer\Exclude()
     * @Serializer\Type("string")
     */
    protected $_api_key_private;
```
Has you can see the api_key_private is exclude from the serializer.

In the controller we use function to update the user :
```php
$user = $this->deserializeEntity($request->request->all(), $user);
```
deserializeEntity update only the entity property they are not excluded.

And in the controller i don't remove a line doing a setApiKeyPrivate function.
Then old test is invalid.
Why is working before, I ignore it. But with my modifications the test fail now and I really don't know why. But i think skip the test is not a better solution.

Rest/Controller/UserController.php
crouillon: I think we cannot check on an empty password.
Prefer to check a boolean field resetPassword: if TRUE, generate a new password for the user and force him to choose a new one a his first connection.

So from the toolbar, the creation form will always send true, and the edit form will propose a checkbox to admin in order to reset or not the user's password

mickaelandrieu:
Thank you for explanation.

For me, it's :+1:

Rest/Controller/UserController.php
```diff
@@ -59,14 +59,13 @@ class UserController extends ARestController
...
     {
-        // TODO
-
-
         if (!$this->isGranted('VIEW', new ObjectIdentity('class', 'BackBee\Security\User'))) {
             throw new AccessDeniedException(sprintf('You are not authorized to view users'));
         }
 
-        return array();
+        $users = $this->getEntityManager()->getRepository('BackBee\Security\User')->findAll();
```
eric-chau: Instead of calling ``::findAll()``, maybe should we paginate result? Cause we won't get any problem as long as there is not alot of users, but what if we do? Anyway it's just a suggestion, clearly we don't need to fix it now but don't forget to create an issue about it.

Rest/Controller/UserController.php
```diff
@@ -120,6 +119,43 @@ public function deleteAction($id)
         return new Response("", 204);
     }
 
+
+    /**
+     * PATCH User
+     *
+     * @param int $id User ID
+     */
+    public function patchAction($id, Request $request)
```
eric-chau: I can't see in ``route.yml`` any route that require this action, is it normal?

Rest/Controller/UserController.php
```diff
@@ -120,6 +119,43 @@ public function deleteAction($id)
...
+    {
+        if (!$this->isGranted('IS_AUTHENTICATED_FULLY')) {
+            throw new AccessDeniedException('You must be authenticated to view users');
+        }
+
+        $user = $this->getApplication()->getBBUserToken()->getUser();
+
+        if ($user->getId() !== $id) {
+            throw new AccessDeniedException('Only user can choose his password');
```
eric-chau: The exception's message should be more specific: "Password can only be changed by its owner." or something like that :P

Rest/Controller/UserController.php
ndufreche: I will change that when I will do the integration of the pagination into the toolbar

Rest/Controller/UserController.php
```diff
@@ -120,6 +119,43 @@ public function deleteAction($id)
...
+        }
+
+        if ($user->getState() === User::PASSWORD_NOT_PICKED) {
+            $this->updateUserPassword($user, $request);
+        } else {
+            $encoderFactory = $this->getContainer()->get('security.context')->getEncoderFactory();
+            $oldPassword = $request->get('old_password', '');
```
eric-chau: Maybe you can throw an exception if ``$oldPassword`` is empty (``=== ''``) so you don't need to run the password encoding process. If i'm right you cannot create a password with an empty string or if it has not alteast 6 characters.

Rest/Controller/UserController.php
eric-chau: Ok =)

Rest/Controller/UserController.php
ndufreche: Yeah, it's not used for the moment. I just do a little anticipation of the backbee/BbCoreJs#164 issue

eric-chau: Ok everything is fine so ^^

Rest/Controller/UserController.php
ndufreche: Yeah that's clear, I have to implement Assert violation style Symfony Response.

Rest/Controller/UserController.php
eric-chau: I think these checks are necessary cause a security context doesn't necessary have an encoder factory like an encoder factory don't necessary has an encoder for the provided user.

Rest/Controller/UserController.php
```diff
@@ -200,22 +236,54 @@ public function postAction(Request $request)
         }
 
         $user = $this->deserializeEntity($request->request->all(), $user);
-
         // handle the password
         if ($request->request->has('password')) {
-            $encoderFactory = $this->getContainer()->get('security.context')->getEncoderFactory();
-            $password = $request->request->get('password');
+           $password = $request->request->get('password');
```
eric-chau: the indentation is not valid =)

Rest/Controller/UserController.php
```diff
@@ -200,22 +236,54 @@ public function postAction(Request $request)
...
+        } elseif ($request->request->has('generate_password') && true === $request->request->get('generate_password')) {
+            $password = substr(hash('sha512', rand()), 0, 6);
+        } else {
+            return new Response('{"errors":{"password": ["Password not provided"]}}', 400, ['Content-Type' => 'application/json']);
```
eric-chau: To be sure to generate a valid json format, you should use ``Symfony\Component\HttpFoundation\JsonResponse`` and writing something like that:
```php
return new JsonResponse([
    'errors' => [
        'password' => ['Password not provided']
    ]
], 400);
```
``JsonResponse`` will call ``json_encode`` on data you passed to its constructor first argument.

Rest/Controller/UserController.php
```diff
@@ -200,22 +236,54 @@ public function postAction(Request $request)
...
+
         $this->getEntityManager()->persist($user);
         $this->getEntityManager()->flush();
 
-        return new Response($this->formatItem($user), 200, array('Content-Type' => 'application/json'));
+        return new Response($this->formatItem($user), 200, ['Content-Type' => 'application/json']);
+    }
+
+    /**
+     * @todo  set minimal password size in secirity config
```
eric-chau: typo, security*

Rest/Controller/UserController.php
```diff
@@ -200,22 +236,54 @@ public function postAction(Request $request)
...
+     * [updateUserPassword description]
+     * @param  User    $user    [description]
+     * @param  Request $request [description]
+     * @return [type]           [description]
+     */
+    private function updateUserPassword(User $user, Request $request)
+    {
+        if ($request->request->get('password') !== $request->request->get('confirm_password')) {
+            throw new AccessDeniedException('password and confirm password are differents');
```
eric-chau: This is not an access denied but a invalid argument exception.

Rest/Controller/UserController.php
```diff
@@ -200,22 +236,54 @@ public function postAction(Request $request)
...
+        }
+        $password = trim($request->request->get('password'));
+        if (strlen($password) < 5) {
+            throw new AccessDeniedException('Password to short.');
```
eric-chau: Same here, ``InvalidArgumentException`` is more appropriate.

Rest/Controller/UserController.php
```diff
@@ -200,22 +236,54 @@ public function postAction(Request $request)
...
+
+        $encoderFactory = $this->getContainer()->get('security.context')->getEncoderFactory();
+
+        if ($encoderFactory && $encoder = $encoderFactory->getEncoder($user)) {
+            $password = $encoder->encodePassword($password, '');
+        }
+
+        $user->setPassword($password);
+        $this->getEntityManager()->persist($user);
+        $this->getEntityManager()->flush();
```
eric-chau: Don't hesistate to use ``$this->getEntityManager()->flush($user);`` instead of ``$this->getEntityManager()->flush();`` cause if there is other modifications managed by application's entity manager these will be saved too, which is probably not what you want to do.

Rest/Controller/UserController.php
ndufreche: This property is not used to be injected into the database but the new password length is tested.
This is like identity check before set the new password, i think the real problem is the implementation, basicly I've to use the security system to re-autenticate the user with the old password.

eric-chau: You're right

mickaelandrieu:
@ndufreche I've relaunched Scrutinizer build. Since your tests are green, you can merge it.

ndufreche:
@mickaelandrieu thanks

eric-chau:
Great, thanks @ndufreche. I merge it.

ndufreche:
\o/

Issue #244, opened 2015-02-12
Uncouple Logger from error handler


mickaelandrieu:
Error Handler and Logger don't need to be coupled.

My final objective is to be able to use the new Symfony varDumper Component as helper and the old (but so good) Debug component as error handler.

mickaelandrieu:
For instance, let's add some weird code into ``FrontController handle()`` source code:

```php
public function handle(Request $request = null, $type = self::MASTER_REQUEST, $catch = true)
    {
        dump(['foo', 'bar', 'baz']);
        throw new \Exception('This is an exception catched by Symfony debugguer and written in log file.');
[...]
```

And take a look at the render in ``debug mode``:

![real_debug_tools](https://cloud.githubusercontent.com/assets/1247388/6168571/b92ef6bc-b2c6-11e4-99ed-4fea1340d8d0.png)

I wanted (also) to move to monolog logger *but* at this point I can't because this enforce me to totaly rewrite some Data collectors and our own profiler: this is out of scope of this ticket.

PR #245, opened 2015-02-12
[RFR] - Uncouple Logger from errorHandler/exceptionHandler

mickaelandrieu:
See #244

This will uncouple error/exception handling from rendering, and also I've added the new [Symfony VarDumper](http://symfony.com/blog/new-in-symfony-2-6-vardumper-component) which is very powerfull to debug, and also it won't break the script.

real_debug_tools


mickaelandrieu:
Thanks @crouillon for review

Issue #250, opened 2015-02-18
Error on autoloading class Backbee\Validator\Rules\Image


fkroockmann:
The autoloader expected class "BackBee\Validator\Rules\Image" to be defined in file "/vendor/backbee/backbee/Validator/Rules/Image.php". The file was found but the class was not in it, the class name or namespace probably has a typo.

PR #251, opened 2015-02-18
#250 fix Error on autoloading class Backbee\Validator\Rules\Image

fkroockmann:
The autoloader expected class "BackBee\Validator\Rules\Image" to be defined in file "/vendor/backbee/backbee/Validator/Rules/Image.php". The file was found but the class was not in it, the class name or namespace probably has a typo.


mickaelandrieu:
Thanks @fkroockmann !

Issue #227, opened 2015-02-02
[rest] Complete rest group controller


ndufreche:
- [ ] A Groups have to return the user they have

mickaelandrieu:
@ndufreche what is the current status of this issue ?

eric-chau:
@ndufreche can we close this issue? It seems completed.

ndufreche:
Yes :)

eric-chau:
Thank you =)

PR #253, opened 2015-02-20
[rest] Complete rest group controller #227

ndufreche:
This is a REST and Entity update of Group. The issue #227 still in working progress.

### DIFF
--- a/Command/Tests/AclLoadCommandTest.php
+++ b/Command/Tests/AclLoadCommandTest.php
@@ -50,10 +50,7 @@ class AclLoadCommandTest extends TestCase
         $this->initAcl();
 
         $superAdminGroup = new Group();
-        $superAdminGroup
-            ->setIdentifier('super_admin')
-            ->setName('Super Admin')
-        ;
+        $superAdminGroup->setName('Super Admin');
         $this->bbapp->getEntityManager()->persist($superAdminGroup);
 
         $this->getBBApp()->getEntityManager()->flush();
@@ -73,7 +70,7 @@ class AclLoadCommandTest extends TestCase
 
         vfsStream::umask(0000);
         $this->_mock_basedir = vfsStream::setup('test_dir', 0777, array(
-            'file.xml' => '
+            'file.yml' => '
 groups:
   super_admin:
     sites:
@@ -85,11 +82,11 @@ groups:
         $commandTester->execute(
             array(
                 'command' => $command->getName(),
-                'file' => 'vfs://test_dir/file.xml',
+                'file' => 'vfs://test_dir/file.yml',
             )
         );
 
-        $this->assertContains('Processing file: vfs://test_dir/file.xml', $commandTester->getDisplay());
+        $this->assertContains('Processing file: vfs://test_dir/file.yml', $commandTester->getDisplay());
     }
 
     /**
--- a/Config/route.yml
+++ b/Config/route.yml
@@ -53,8 +53,8 @@ bb.rest.user.get:
         _action: getAction
         _controller: BackBee\Rest\Controller\UserController
     requirements:
-        version: \d
-        id: \d
+        version: \d+
+        id: \d+
         _method: GET
 
 bb.rest.user.post:
@@ -63,7 +63,7 @@ bb.rest.user.post:
         _action: postAction
         _controller: BackBee\Rest\Controller\UserController
     requirements:
-        version: \d
+        version: \d+
         _method: POST
 
 bb.rest.user.put:
@@ -72,8 +72,8 @@ bb.rest.user.put:
         _action: putAction
         _controller: BackBee\Rest\Controller\UserController
     requirements:
-        version: \d
-        id: \d
+        version: \d+
+        id: \d+
         _method: PUT
 
 bb.rest.user.delete:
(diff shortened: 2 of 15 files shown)


Rest/Controller/GroupController.php
```diff
@@ -42,32 +45,27 @@ class GroupController extends ARestController
...
     public function getCollectionAction(Request $request)
     {
+        if ($request->request->has('site_uid')) {
+            $site_uid = $request->request->get('site_uid');
+            $this->checkSiteUid($site_uid);
+        } else {
+            $site_uid = $this->getApplication()->getSite()->getUid();
+        }
+
         $qb = $this->getEntityManager()->createQueryBuilder()
```
mickaelandrieu: I don't like the queries to be created inside the controller.

Why not use a repository ?

Rest/Controller/GroupController.php
```diff
@@ -124,53 +115,52 @@ public function putAction(Group $group, Request $request)
...
         $this->getEntityManager()->flush();
 
-        return new Response($this->formatItem($group), 200, array('Content-Type' => 'application/json'));
+        return new Response($this->formatItem($group), 200, ['Content-Type' => 'application/json']);
+    }
+
+    private function checkSiteUid($site_uid)
+    {
+        $site = $this->getEntityManager()->find('BackBee\Site\Site', $site_uid);
+        $validator = Validation::createValidator();
```
mickaelandrieu: the best we can do here, for the future will be to add a new helper inside BackBee\Controller called ``validateValue($value, ['constraintClass1', 'constraintClass2','..']`` which return a ConstraintViolationList.

Then, we may have another helper called ``validateValueOrThrow($value, [])`` in charge of throwing the correct exception. This way your controller stay as thin as possible.

Rest/Controller/SiteController.php
```diff
@@ -24,6 +24,7 @@
 namespace BackBee\Rest\Controller;
 
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException;
```
mickaelandrieu: where did you need it ?

Security/Acl/Loader/YmlLoader.php
```diff
@@ -65,11 +65,10 @@ public function load($aclData)
         }
 
         foreach ($grid['groups'] as $group_identifier => $rights) {
-            if (null === $group = $this->em->getRepository('BackBee\Security\Group')->findOneBy(array('_identifier' => $group_identifier))) {
+            if (null === $group = $this->em->getRepository('BackBee\Security\Group')->findOneBy(array('_name' => $group_identifier))) {
                 // ensure group exists
                 $group = new \BackBee\Security\Group();
-                $group->setIdentifier($group_identifier)
-                        ->setName($group_identifier);
+                $group->setName($group_identifier);
```
mickaelandrieu: Sounds weird.

I suggest:
```php
foreach ($grid['groups'] as $group_name => $rights)
```

Tests/TestCase.php
```diff
@@ -281,13 +281,14 @@ protected function createAuthUser($groupId, $roles = array())
 
         $group = $this->getBBApp()->getEntityManager()
             ->getRepository('BackBee\Security\Group')
-            ->findOneBy(array('_identifier' => $groupId))
+            ->findOneBy(array('_name' => $groupId))
```
mickaelandrieu: sounds weird too. A name is not supposed to be unique.

Rest/Controller/SiteController.php
ndufreche: Because it was used after, in the old code.

Tests/TestCase.php
ndufreche: It is for create easily a group in test.

Command/Tests/AclLoadCommandTest.php
```diff
@@ -50,9 +50,7 @@ protected function setUp()
         $this->initAcl();
 
         $superAdminGroup = new Group();
-        $superAdminGroup
-            ->setIdentifier('super_admin')
-            ->setName('Super Admin')
+        $superAdminGroup->setName('Super Admin')
         ;
```
eric-chau: you can move this to the line above =)

Config/route.yml
```diff
@@ -97,15 +97,25 @@ bb.rest.user.get_collection:
...
 bb.rest.group.get:
-    pattern: /rest/{version}/group/{uid}
+    pattern: /rest/{version}/group/{id}
     defaults:
         _action:     getAction
         _controller: BackBee\Rest\Controller\GroupController
     requirements:
         version: \d
-        uid: \w{32}
+        id: \d
```
eric-chau: you need to add a ``+``, ``\d+``. I think you don't meet any issue until now but maybe it's because your id is only composed by one digit. I'm not sure that ``10`` will work.

Rest/Controller/GroupController.php
eric-chau: I was about to say the same thing.

Rest/Controller/GroupController.php
ndufreche: Yeah it's good Idea.
I think, a specific issue this needed for that.

mickaelandrieu: true.

mickaelandrieu:
Thank you @ndufreche

Issue #256, opened 2015-02-24
[Config] update the way we compute keys in Persistor


eric-chau:
``BackBee\Config\Persistor\File`` and ``BackBee\Config\Persistor\Registry`` use wrong key, it must be fixed and use ``BackBee\Bundle\BundleLoader::getServiceIdByBaseDir`` for more consistency.

PR #257, opened 2015-02-24
[RFR] [Config] #256 + updated the way keys are computed in File and Registry persistor

eric-chau:
More details at #256.


mickaelandrieu:
:+1:

PR #260, opened 2015-02-25
#259 - Add UserControllerTest::getCollectionAction()

crouillon:


eric-chau:
Be used to provide a description please so we can all understand what happens.

Rest/Tests/Controller/UserControllerTest.php
```diff
@@ -118,6 +118,23 @@ public function testGetAction()
...
+     */
+    public function testGetCollectionAction()
+    {
+        $controller = $this->getController();
+
+        // no filters
+        $response = $this->getBBApp()->getController()->handle(new Request(array(), array(
+        ), array(
+            '_action' => 'getCollectionAction',
+            '_controller' => 'BackBee\Rest\Controller\UserController',), array(), array(), array('REQUEST_URI' => '/rest/1/test/')));
```
mickaelandrieu: are you sure ?

crouillon:
See with @ndufreche, we need to test the return type of UserController::getCollectionAction() to be sure to get an array and not an object

eric-chau:
Ok thanks at you both :+1:

PR #261, opened 2015-02-26
[RFR] Introduced a new way to execute and configure error pages

mickaelandrieu:
I moved the exception management to a ``kernel.exception`` listener.

Also, imo we should be able to override the error templates because our clients always want a customised error page, with their own content and dependencies.

This is an actual POC, but I need a review and advices to improve my commits.

Finally, how can I test it ?

### DIFF
--- a/Event/Listener/ExceptionListener.php
+++ b/Event/Listener/ExceptionListener.php
@@ -1,0 +1,96 @@
+<?php
+
+/*
+ * Copyright (c) 2011-2015 Lp digital system
+ *
+ * This file is part of BackBee.
+ *
+ * BackBee5 is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * BackBee is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with BackBee. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * @author Charles Rouillon <charles.rouillon@lp-digital.fr>
+ */
+
+namespace BackBee\Event\Listener;
+
+use BackBee\BBApplication;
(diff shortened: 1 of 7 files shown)


Event/Listener/ExceptionListener.php
```diff
@@ -0,0 +1,92 @@
...
+     * @var BackBee\Renderer
+     */
+    private $renderer;
+
+    /**
+     * @var Symfony\Component\HttpFoundation\Response
+     */
+    private $response;
+
+    public function __construct(BBApplication $bbapp)
```
fkroockmann: $bbapp -> $application (better :))

Event/Listener/ExceptionListener.php
```diff
@@ -0,0 +1,92 @@
...
+    {
+        $this->bbapp = $bbapp;
```
fkroockmann: same here

Event/Listener/ExceptionListener.php
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    /**
+     * Returns the path of the template for the selected HTTP Code error status
+     *
+     * @input string $error 404|500
+     *
+     * @return string
+     */
+    private function getErrorTemplate($error)
```
fkroockmann: $errorName ?

Resources/layouts/error/500.phtml
```diff
@@ -0,0 +1,33 @@
+<!DOCTYPE html>
```
fkroockmann: French words ?

Resources/layouts/error/404.phtml
```diff
@@ -0,0 +1,33 @@
+<!DOCTYPE html>
```
fkroockmann: French words ?

mickaelandrieu: yeah :/ as before. Do we have a translator in PHP side ?

fkroockmann: I Don't know :/

mickaelandrieu: Ok for translations but in another PR as I want to improve the system to allow the choice of format (xliff, yaml, php array, etc).

For now, all templates are in french, so *for consistency* I provide french templates.

mickaelandrieu:
ping @eric-chau. This is now done, translations done will be done in another PR ( when I will make the translator able to use every loader from Symfony translator component).

Issue #279, opened 2015-03-03
Fatal error on bb5content::_setElementName(AClassContent $content)


crouillon:
Because in some case $content is not instance of AClassContent but Page

PR #281, opened 2015-03-03
[Renderer] #279 - change bb5content::_setElementName to accept all object but return an intance of AClassContent

adloiseau:


mickaelandrieu:
Please, don't merge without a review.
Also the description has nothing to to with the changes done in this contribution.

crouillon:
I reviewed it before I merged
And indeed there's missing a negation in title, I change it

PR #286, opened 2015-03-17
[wip] finalising and debuging rest user and group

ndufreche:
- [x] Create user patch actions
  - [x] Patch identity
  - [x] Patch password
  - [x] Patch groups
  - [x] Patch status
- [x] Complete user getCollection action
  - [x] getCollection by group
- [x] Create getCurrent user action 
- [x] Update Group entity
  - [x] Serialize users
  - [x] Rename setUser function in addUser
- [x] Update User entity
  - [x] Serialize groups
  - [x] Normalize the entity (serializer annotations)

### DIFF
--- a/Config/route.yml
+++ b/Config/route.yml
@@ -58,6 +58,15 @@ bb.rest.resource.upload:
         _method: POST
 
 # User Rest API
+bb.rest.user.current:
+    pattern: /rest/{version}/user/current
+    defaults:
+        _action: getCurrentAction
+        _controller: BackBee\Rest\Controller\UserController
+    requirements:
+        version: \d+
+        _method: GET
+
 bb.rest.user.get:
     pattern: /rest/{version}/user/{id}
     defaults:
@@ -86,6 +95,16 @@ bb.rest.user.put:
         version: \d+
         id: \d+
         _method: PUT
+
+bb.rest.user.patch:
+    pattern: /rest/{version}/user/{id}
+    defaults:
+        _action: patchAction
+        _controller: BackBee\Rest\Controller\UserController
+    requirements:
+        version: \d+
+        id: \d+
+        _method: PATCH
 
 bb.rest.user.delete:
     pattern: /rest/{version}/user/{id}
--- a/Rest/Controller/UserController.php
+++ b/Rest/Controller/UserController.php
@@ -23,14 +23,22 @@
 
 namespace BackBee\Rest\Controller;
 
+use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Exception\ConflictHttpException;
 use Symfony\Component\Security\Acl\Domain\ObjectIdentity;
 use Symfony\Component\Security\Core\Exception\AccessDeniedException;
 use Symfony\Component\Validator\Constraints as Assert;
+use Symfony\Component\Validator\Validation;
 
 use BackBee\Rest\Controller\Annotations as Rest;
+use BackBee\Rest\Exception\ValidationException;
+use BackBee\Rest\Patcher\EntityPatcher;
+use BackBee\Rest\Patcher\Exception\InvalidOperationSyntaxException;
+use BackBee\Rest\Patcher\Exception\UnauthorizedPatchOperationException;
+use BackBee\Rest\Patcher\OperationSyntaxValidator;
+use BackBee\Rest\Patcher\RightManager;
 use BackBee\Security\User;
 
 /**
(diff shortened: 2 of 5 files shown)


Rest/Controller/UserController.php
```diff
@@ -59,16 +67,41 @@ class UserController extends ARestController
...
         }
 
-        $users = $this->getEntityManager()->getRepository('BackBee\Security\User')->findAll();
+        $group = $request->query->get('groups', null);
+
+        if ($group !== null) {
+            $group = $this->getEntityManager()->find('BackBee\Security\Group', $group);
+            $users = $group->getUsers();
+        } else {
+            $users = $this->getEntityManager()->getRepository(get_class($this->getUser()))->findAll();
```
eric-chau: You should use ``BackBee\Security\User`` instead of ``get_class($this->getUser())`` cause sometimes you might get the entity proxy classname instead of the real one.

Config/route.yml
```diff
@@ -47,6 +47,15 @@ bb.images:
 # BackBee REST API
 
 # User Rest API
+bb.rest.user.current:
+    pattern: /rest/{version}/user/current
+    defaults:
+        _action: getCurrent
```
eric-chau: Rename it to ``getCurrentAction``, it's a best pratice to suffix every action by ``Action``.

Rest/Controller/UserController.php
```diff
@@ -59,16 +67,41 @@ class UserController extends ARestController
...
     /**
+     * GET current User
+     */
+    public function getCurrent()
+    {
+        if (!$this->isGranted('IS_AUTHENTICATED_FULLY')) {
+            throw new AccessDeniedException('You must be authenticated to access');
+        }
+
+        $user = $this->getEntityManager()->find(get_class($this->getUser()), $this->getUser()->getId());
```
eric-chau: Same comment than above about using ``get_class()``.

Rest/Controller/UserController.php
```diff
@@ -79,7 +112,7 @@ public function getAction($id)
             throw new AccessDeniedException('You must be authenticated to delete users');
         }
 
-        $user = $this->getEntityManager()->getRepository('BackBee\Security\User')->find($id);
+        $user = $this->getEntityManager()->find(get_class($this->getUser()), $id);
```
eric-chau: Same here, don't use ``get_class()``. Also it implies that you have to execute 2 functions to get the namespace (that you're not sure it's the right one) while you already know what it is.

Rest/Controller/UserController.php
```diff
@@ -103,7 +136,11 @@ public function deleteAction($id)
             throw new AccessDeniedException('You must be authenticated to delete users');
         }
 
-        $user = $this->getEntityManager()->getRepository('BackBee\Security\User')->find($id);
+        if ($this->getUser()->getId() == $id) {
```
eric-chau: It's a best pratice to use ``===``, I suggest you to do something like that: ``intval($id) === $this->getUser()->getId()``.

Rest/Controller/UserController.php
```diff
@@ -103,7 +136,11 @@ public function deleteAction($id)
...
+            throw new AccessDeniedException('You can remove the user of your current session.');
+        }
+
+        $user = $this->getEntityManager()->find(get_class($this->getUser()), $id);
```
eric-chau: Same here about the use of ``get_class()`` =)

Rest/Controller/UserController.php
```diff
@@ -232,7 +239,7 @@ public function postAction(Request $request)
             throw new AccessDeniedException('You must be authenticated to view users');
         }
 
-        $userExists = $this->getApplication()->getEntityManager()->getRepository('BackBee\Security\User')->findBy(array('_login' => $request->request->get('login')));
+        $userExists = $this->getApplication()->getEntityManager()->getRepository(get_class($this->getUser()))->findBy(array('_login' => $request->request->get('login')));
```
eric-chau: Don't hesitate to write your instruction on several lines so you can respect the 120 characters max per line.

Rest/Controller/UserController.php
ndufreche: No because BackBee\Security\User can be extended to add new parameters for custom products, and normally you get the proxy if you retrieve an object with doctrine, here, is the user from the token, it safely signed.

eric-chau: Ok then =)

mickaelandrieu:
This sounds good, if you have time can you add minimal tests for PATCH method ?

eric-chau:
:+1:, thanks for your work, yeah get some times to add some tests about your patch method =)

Rest/Controller/UserController.php
```diff
@@ -273,22 +282,132 @@ public function postAction(Request $request)
...
+        $violations = $validator->validateValue($operations['email'], $constraint);
+        if (count($violations) > 0) {
+            throw new ValidationException($violations);
+        }
+
+        $user = $this->getEntityManager()->find(get_class($this->getUser()), $id);
+
+        $user->setFirstname($operations['firstname']);
+        $user->setLastname($operations['lastname']);
+        $user->setEmail($operations['email']);
```
ndufreche: @eric-chau What do you think about iterate on $operations to do a dynamic assignment ?

ndufreche:
I'm agree to do some test on the patch action but not on this PR, because clement need to test it tomorrow.
I will start a new PR tomorrow with the patch action tests.
This is ok for you @eric-chau and @mickaelandrieu 
I remove the getCollection for the search engine, I will complete in another PR.

Rest/Controller/UserController.php
eric-chau: It's supposed to already work this way. Look at ``BackBee\Rest\Controller\PageController::patchAction``.

mickaelandrieu:
Ok for me.

Rest/Controller/UserController.php
ndufreche: Ok I will upgrade that, in the same time how I will do the tests

PR #288, opened 2015-03-23
[REST] [Resource] + add upload from base64 src

fkroockmann:
- Split main function
- Add upload from base64 src


eric-chau:
Add some comments please to your function.

Issue #101, opened 2014-12-10
[Insight] Unused method, property, variable or parameter - in Bundle/Registry/Builder.php, line 39


mickaelandrieu:
**[in Bundle/Registry/Builder.php, line 39](https://github.com/backbee/BackBee/blob/master/Bundle/Registry/Builder.php#L39)**

> This `em` attribute is declared but never used. You should remove it.

```php
class Builder
{
    /**
     * @var EntityManager
     */
    private $em;
    /**
     * @var boolean
     */
    private $is_registry_entity = false;
    /**
```

---

_Posted from [SensioLabsInsight](https://insight.sensiolabs.com/projects/c9dd6e24-288f-41a9-8400-82e5fae6015b/analyses/57)_

PR #293, opened 2015-03-26
backbee/backbee-standard#101 : Prepare 1.0.0

nicolas-bastien:


mickaelandrieu:
Same here: what is the correct way to handle with releases in your opinion ?

mickaelandrieu:
Thank you @nicolas-bastien

PR #311, opened 2015-04-07
[Event] + change the event on flush to on pre flush for element file into RevisionListener

fkroockmann:
Bad use of event cause of change the data into a flush event instead of a pre flush event.

### DIFF
--- a/Event/Listener/RevisionListener.php
+++ b/Event/Listener/RevisionListener.php
@@ -112,7 +112,7 @@ class RevisionListener
         }
     }
 
-    public static function onFlushElementFile(Event $event)
+    public static function onPreFlushElementFile(Event $event)
     {
         $application = $event->getApplication();
         $em = $application->getEntityManager();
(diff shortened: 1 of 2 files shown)


Event/Listener/RevisionListener.php
```diff
@@ -120,10 +120,15 @@ public static function onFlushElementFile(Event $event)
 
         $revision = $event->getTarget();
         $content = $revision->getContent();
-
+        
+        if ($uow->isScheduledForDelete($content)) {
+            return;
```
eric-chau: You have to check if there is a file that exists at its ``path`` property location cause here it's a hard delete of the entity (not a simple change of its ``state``).

fkroockmann: What do you think @crouillon, we delete the file or we keep it ?

fkroockmann:
@eric-chau, i create a issue for add a delete treatment.
https://github.com/backbee/BackBee/issues/312

eric-chau:
ok @fkroockmann, atleast we won't forget to do it later. Thanks.

PR #313, opened 2015-04-07
[Renderer] + don't force to set timezone/culture/format in datehelper

fkroockmann:
- Use default timezone/culture if is not set into config
- Use a format "%Y-%m-%d at %H:%M" if is not set into config
- In initTimzone -> verify if config is an array before use array_key_exists function


mickaelandrieu:
thank you @fkroockmann  !

PR #315, opened 2015-04-08
creating isDebug helper

ndufreche:
creating isDebug helper

### DIFF
--- a/Renderer/Helper/isDebug.php
+++ b/Renderer/Helper/isDebug.php
@@ -1,0 +1,42 @@
+<?php
+
+/*
+ * Copyright (c) 2011-2013 Lp digital system
+ *
+ * This file is part of BackBee.
+ *
+ * BackBee is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


mickaelandrieu:
Before all new additions, I want an installable package for 1.0.*.

I will only accepts bug fix since then /c @crouillon .

So, please focus on this (important) task.

ndufreche:
Is an important add, it is for identify what type of javascript we have to use in toolbar-bundle and it's you @mickaelandrieu who say to put it in the php core...
see backbee/ToolbarBundle#43

mickaelandrieu:
yes, I think this need to be pushed in BackBee, but I need an installable package for 1.0 since 2 weeks now, so I prefer you focus on it if you don't mind.

I can confirm this will be merged then.

ndufreche:
Yes but I need it into backbee/ToolbarBundle#44

mickaelandrieu:
Merged.

Thanks @ndufreche

mickaelandrieu:
@ndufreche your next PR will be on backbee-standard to push bb-core-dist.
Until then, I will only accept bug fixes: consider PHP projects frozen.

PR #318, opened 2015-04-21
[Rest] Set null the redirect if empty on PageController

fkroockmann:
Currently redirect is always set as empty

### DIFF
--- a/Rest/Controller/PageController.php
+++ b/Rest/Controller/PageController.php
@@ -213,11 +213,14 @@ class PageController extends AbstractRestController
             $builder->setSite($this->getApplication()->getSite());
         }
 
+        $requestRedirect = $request->request->get('redirect');
+        $redirect = ($requestRedirect === '' || $requestRedirect === null) ? null : $requestRedirect;
+
         $builder->setTitle($request->request->get('title'));
         $builder->setUrl($request->request->get('url', null));
         $builder->setState($request->request->get('state'));
         $builder->setTarget($request->request->get('target'));
-        $builder->setRedirect($request->request->get('redirect'));
+        $builder->setRedirect($redirect);
         $builder->setAltTitle($request->request->get('alttitle'));
         $builder->setPublishing(
             null !== $request->request->get('publishing')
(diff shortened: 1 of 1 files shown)


Rest/Controller/PageController.php
```diff
@@ -213,11 +213,14 @@ public function postAction(Layout $layout, Request $request, Page $parent = null
             $builder->setSite($this->getApplication()->getSite());
         }
 
+        $requestRedirect = $request->request->get('redirect');
```
mickaelandrieu:
```php
$requestRedirect = $request->request->get('redirect', null);
```

Rest/Controller/PageController.php
```diff
@@ -295,11 +298,15 @@ public function putAction(Page $page, Layout $layout, Request $request)
         $page->setLayout($layout);
         $this->trySetPageWorkflowState($page, $this->getEntityFromAttributes('workflow'));
 
+
+        $requestRedirect = $request->request->get('redirect');
+        $redirect = ($requestRedirect === '' || $requestRedirect === null) ? null : $requestRedirect;
```
mickaelandrieu:
```php
$requestRedirect = $request->request->get('redirect', null);
```

Rest/Controller/PageController.php
fkroockmann: This test will tell if the parameter does not exist, you set a value to null. Except that the current problem is that the "redirect" is knowledgeable but it is an empty string and an empty string is not a url ...

Rest/Controller/PageController.php
fkroockmann: This test will tell if the parameter does not exist, you set a value to null. Except that the current problem is that the "redirect" is knowledgeable but it is an empty string and an empty string is not a url ...

mickaelandrieu: ok, i merge

Rest/Controller/PageController.php
eric-chau: Anyway ``$requestRedirect = $request->request->get('redirect', null);`` is the default behavior, check the api.

fkroockmann: What api ?

fkroockmann: Yes, it's a js problem

eric-chau: ``Symfony\Component\HttpFoundation\ParameterBag::get``. It returns ``null`` only when the key does not exist (here in post data), but I guess you submit it in your form (there is an ``input[type="text"]`` named redirect) and empty is translated to empty string, not null which is logical.

fkroockmann: Yes :), that's what I meant to mickae.
But the problem is not in php, i will revert this PR.

eric-chau: @fkroockmann, your PR still legit, we can also check it in PHP side. But maybe move the test into ``BackBee\NestedNode\Page::setRedirect`` by something like this:

```php
    public function setRedirect($redirect = null)
    {
        $redirect = trim($redirect);
        $this->_redirect = empty($redirect) ? null : $redirect;
        return $this;
    }
```

PR #320, opened 2015-04-21
[NestedNode] Fix the default classcontent state when user create a page

fkroockmann:
Currently the state of default classcontent is "new" instead of "normal";


crouillon:
It seems to be the correct state: while creating a new page its contents have to be committed before being "normal"

fkroockmann:
After creating a page, there is nothing to be validated , the bug is on the recovery of validating content ?

Issue #228, opened 2015-02-03
[REST] [Page] missing data


fkroockmann:
On a getAction, the modified and created data are not retrieved

mickaelandrieu:
Hi @fkroockmann, can you create a minimal test which won't pass because the actual behavior is not what we expect ?

mickaelandrieu:
@fkroockmann can you confirm bug ?

eric-chau:
This is not a bug but I confirm that the created and modified data are missing.

PR #334, opened 2015-05-02
[RFR] [NestedNode] added two methods to Page entity #228

eric-chau:
Added two methods ``Page::getCreatedTimestamp`` and ``Page::getModifiedTimestamp`` and annoted as JMS virtual property. It solves issue #228 by adding ``created`` and ``modified`` data to page rest api.


mickaelandrieu:
thank you @eric-chau

Issue #218, opened 2015-01-27
Fixed doctrine mapping issues


mickaelandrieu:
Manager default
Class	Mapping errors
BackBuilder\NestedNode\Page	
The field BackBuilder\NestedNode\Page#_mainsection is on the inverse side of a bi-directional relationship, but the specified mappedBy association on the target-entity BackBuilder\NestedNode\Section#_page does not contain the required 'inversedBy=_mainsection' attribute.
Field '_section' in class 'BackBuilder\NestedNode\Page' must be private or protected. Public fields may break lazy-loading.
Field '_mainsection' in class 'BackBuilder\NestedNode\Page' must be private or protected. Public fields may break lazy-loading.
Field '_subcontent' in class 'BackBuilder\ClassContent\AClassContent' must be private or protected. Public fields may break lazy-loading.
BackBuilder\ClassContent\ContentSet	
Field '_subcontent' in class 'BackBuilder\ClassContent\ContentSet' must be private or protected. Public fields may break lazy-loading.
BackBuilder\ClassContent\ContentSet	
Field '_subcontent' in class 'BackBuilder\ClassContent\ContentSet' must be private or protected. Public fields may break lazy-loading.
BackBuilder\ClassContent\Element\file	
Field '_subcontent' in class 'BackBuilder\ClassContent\Element\file' must be private or protected. Public fields may break lazy-loading.
BackBuilder\ClassContent\article	
Field '_subcontent' in class 'BackBuilder\ClassContent\article' must be private or protected. Public fields may break lazy-loading.
BackBuilder\ClassContent\article\body	
Field '_subcontent' in class 'BackBuilder\ClassContent\article\body' must be private or protected. Public fields may break lazy-loading.
BackBuilder\ClassContent\Element\keyword	
Field '_subcontent' in class 'BackBuilder\ClassContent\Element\keyword' must be private or protected. Public fields may break lazy-loading.
BackBuilder\ClassContent\Element\text	
Field '_subcontent' in class 'BackBuilder\ClassContent\Element\text' must be private or protected. Public fields may break lazy-loading.
BackBuilder\ClassContent\Media\image	
Field '_subcontent' in class 'BackBuilder\ClassContent\Media\image' must be private or protected. Public fields may break lazy-loading.
BackBuilder\ClassContent\article\related	
Field '_subcontent' in class 'BackBuilder\ClassContent\article\related' must be private or protected. Public fields may break lazy-loading.
BackBuilder\ClassContent\Element\image	
[...]

PR #337, opened 2015-05-03
[RFR] [ClassContent] [Workflow] fixed Doctrine mapping errors in AbstractClassContent and State #218

eric-chau:
Changed visibility of ``BackBee\ClassContent\AbstractClassContent::$_subcontent`` attribute from ``public`` to ``protected`` otherwise it can break Doctrine lazy-load mechanism.

Added ``inversedBy`` property to ``BackBee\Workflow\State::$_layout`` attribute annotation.

More details on #218.


mickaelandrieu:
thanks !

PR #338, opened 2015-05-03
updated BBApplication by removing some useless attributes and methods and updating coding style

eric-chau:
Cleaned up ``BackBee\BBApplication``:

- removed some unused attributes or methods
- updated methods and attributes names by removing ``_`` prefix
- respecting limit of 120 characters max by line


mickaelandrieu:
:+1:  thank you @eric-chau

PR #343, opened 2015-05-05
[ClassContent] updated ClassContentManager::findOneByTypeAndUid and ClassContentController::getAction

eric-chau:
Updated ``BackBee\ClassContent\ClassContentManager::findOneByTypeAndUid`` to not trying to set draft if content is null.
Updated ``BackBee\Rest\Controller\ClassContentController::getAction`` to throw exception on not found content.

### DIFF
--- a/ClassContent/AbstractContent.php
+++ b/ClassContent/AbstractContent.php
@@ -24,6 +24,7 @@
 namespace BackBee\ClassContent;
 
 use BackBee\AutoLoader\Exception\ClassNotFoundException;
+use BackBee\ClassContent\Exception\InvalidContentTypeException;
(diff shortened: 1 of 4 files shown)


mickaelandrieu:
@eric-chau is this ready for merge ? (maybe some tests are required to check this update actualy do the job ? )

eric-chau:
@mickaelandrieu I was about to resolve #344 into this PR. I will try to add tests too.

eric-chau:
I will complete this PR soonly, we won't notice any error until we provide an invalid classcontent's uid so I have to fix it to also handle error cases.

eric-chau:
ping @mickaelandrieu, this PR is over and resolve issue #344.

PR #345, opened 2015-05-05
Fixed url unicity from urlGenerator

mickaelandrieu:
ping @eric-chau: I'm not sure of my patch => what do you think ?

refs: https://github.com/backbee/backbee-standard/issues/113


Rewriting/UrlGenerator.php
```diff
@@ -277,9 +277,8 @@ private function checkUniqueness(Page $page, &$url)
             ;
         } else {
             $existings = $this->application->getEntityManager()->getConnection()->executeQuery(
-                'SELECT uid FROM page WHERE `root_uid` = :root AND url REGEXP :regex',
```
eric-chau: You cannot remove the test about ``root_uid``, cause each sites has its page tree and if we do so we won't be able to have 2 same URIs on two differents sites.

Rest/Controller/PageController.php
```diff
@@ -436,6 +436,13 @@ public function cloneAction(Page $source, Page $parent = null, $sibling = null,
             $this->getApplication()->getBBUserToken()
         );
 
+        /**
+         * and magic happens
```
mickaelandrieu: ping @eric-chau where do I have to put this lines ? you disagree to let them on controller, but imo (and keeping in mind we will refactor this in months) it's better to see what happens and why.

mickaelandrieu:
closed in favor of https://github.com/backbee/BackBee/pull/350

Issue #324, opened 2015-04-23
Default classcontent is not savable


fkroockmann:
When you create a page with a template having a default content, the content has a "new" status and therefore is not displayed in " not connected" mod . It is also impossible to commit.

fkroockmann:
@mickaelandrieu can you resolve this plz ? :)

eric-chau:
@fkroockmann, this is caused by the fact that a default content created by page and its layout is not linked to any draft. And the classcontent REST API only returns contents that have draft for the current user. It won't be that easy to fix it properly.

PR #346, opened 2015-05-06
[ClassContent] added listener for 'nestednode.page.onflush' event #324

eric-chau:
Added ``BackBee\Event\Listener\ClassContentListener::onFlushPage`` to catch everytime we are persisting a new page event. We are looking for any default classcontent (when the main node is the current page)to create its draft. We also apply the same logic to the default content's subcontents.

This is useful to allow commit of default classcontent with ClassContent REST API.


mickaelandrieu:
thank you @eric-chau

PR #349, opened 2015-05-06
Dispatch rest.user.creation

ndufreche:
We need this specific event to not send email to each user creation ans send this email only if is created in Rest context.


Rest/Controller/UserController.php
```diff
@@ -286,6 +287,10 @@ public function postAction(Request $request)
         $this->getEntityManager()->persist($user);
         $this->getEntityManager()->flush($user);
 
+
+        $event = new Event($user);
+        $this->getApplication()->getEventDispatcher()->dispatch('rest.user.creation', $event);
```
mickaelandrieu: :+1:

mickaelandrieu:
thanks, we should provide a lot of events to make our application more flexible.

PR #351, opened 2015-05-06
[ClassContent] updated RevisionListener::onFlushElementFile

eric-chau:
Updated ``BackBee\Event\Listener\RevisionListener::onFlushElementFile`` to listen ``revision.onflush`` instead of ``revision.preflush``.

### DIFF
--- a/ClassContent/ClassContentManager.php
+++ b/ClassContent/ClassContentManager.php
@@ -71,7 +71,7 @@ class ClassContentManager
     }
 
     /**
-     * Updates the content which provided data.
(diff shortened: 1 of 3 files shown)


mickaelandrieu:
Yes, but why ? Just for understand the process :)

eric-chau:
Yeah you're right. Everything is not clear but when we listen to ``preFlush`` event it's often not triggered while the ``onFlush`` event is always triggered. We also find out other cases about developers that met same issue.

PR #353, opened 2015-05-06
[ClassContent] stop RevisionListener::onFlushElementFile execution if the file's path is not valid

eric-chau:
The move of the file has to be done only in case that the file's path is valid, else it will throw an exception.


mickaelandrieu:
@eric-chau can you put this after the ElementFile check ? We avoid to create 2 variables :)

eric-chau:
yes you're right.

PR #356, opened 2015-05-07
[NestedNode] updated Page::getStateLabel to return right label for offline state

eric-chau:
Aims to fix [backbee/bb-core-js issue #470](https://github.com/backbee/BbCoreJs/issues/470).

This has to be done this way cause ``&`` on 0 will always return 0, that why there was a bug about offline and hidden state label.


mickaelandrieu:
:+1:

Issue #352, opened 2015-05-06
[NestedNode] rewrite page url while it is not online


eric-chau:
We have two issues:

- if page has an URL RewritingListener won't be apply and it causes problem to cloning page
- page url is rewrited only if the page's status change to online

eric-chau:
This issue is resolved by PR #359.

PR #359, opened 2015-05-08
[Rewriting] updated the way we rewrite page's url

eric-chau:
More details on issue #352. The process is also clearer. Also updated coding style.


eric-chau:
At moment do not merge it.

eric-chau:
@mickaelandrieu, @crouillon, I finally ended the ``page's url rewriting process`` and added tests.

mickaelandrieu:
@eric-chau can I merge it ? this sounds good.

eric-chau:
I want to do some more checks.

eric-chau:
And the issue about PHP memory limit still not resolved. It is green cause this problem doesn't occur everytimes so i spammed the build until it pass for each PHP version. :/

eric-chau:
Well, my last commit fixed the issue about PHP memory limit, I simply splitted phpunit's tests to run a new php's script for each component.

PR #360, opened 2015-05-08
Replaced reference to 'AClassContent' by 'AbstractClassContent'

eric-chau:
Some references to ``BackBee\ClassContent\AClassContent`` still remain so I cleaned it. Also Updated some classes coding style.


mickaelandrieu:
@eric-chau let me time to review before merge please, don't hesitate to call me ;)

eric-chau:
Yeah I normally do but here is just change about comment block. Sorry!

PR #362, opened 2015-05-10
[WIP] [Cache] rework of CacheListener

eric-chau:
Currently, the strategy to remove content and page cache is not complete.

### DIFF
--- a/Cache/DAO/Cache.php
+++ b/Cache/DAO/Cache.php
@@ -253,20 +253,22 @@ class Cache extends AbstractExtendedCache
     {
         $tags = (array) $tag;
 
-        if (0 == count($tags)) {
+        if (0 === count($tags)) {
             return false;
         }
 
         try {
             $this->_repository
-                    ->createQueryBuilder('c')
-                    ->delete()
-                    ->where('c._tag IN (:tags)')
-                    ->setParameters(array('tags' => $this->_getContextualTags($tags)))
-                    ->getQuery()
-                    ->execute();
+                ->createQueryBuilder('c')
+                ->delete()
+                ->where('c._tag IN (:tags)')
+                ->setParameters(array('tags' => $this->_getContextualTags($tags)))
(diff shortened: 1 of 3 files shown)


Cache/DAO/Cache.php
```diff
@@ -253,20 +253,22 @@ public function removeByTag($tag)
...
+                ->createQueryBuilder('c')
+                ->delete()
+                ->where('c._tag IN (:tags)')
+                ->setParameters(array('tags' => $this->_getContextualTags($tags)))
+                ->getQuery()
+                ->execute()
+            ;
             $this->_resetCacheEntity();
         } catch (\Exception $e) {
+            die('ERROR!');
```
mickaelandrieu: die ? :p

ClassContent/Repository/IndexationRepository.php
```diff
@@ -352,53 +338,57 @@ public function getDescendantsContentUids($contents)
...
-                ->createQueryBuilder()
-                ->select('c.node_uid')
-                ->from($meta->getTableName(), 'c');
-
-        $q->andWhere('c.'.$meta->getColumnName('_uid').' IN (:ids)')
-              ->setParameter('ids', $content_uids);
-
-        return (false === empty($content_uids)) ? array_unique($q->execute()->fetchAll(\PDO::FETCH_COLUMN)) : array();
+            ->createQueryBuilder()
+            ->select('c.node_uid')
```
crouillon: What do you think about to generalize the use of metadata instead of final column name?
select('c.'.$meta->getColumnName('_mainnode'))
I think it's a good practice to avoid name defined in annotations from code

eric-chau: Yes I think this is a good idea so we don't have to update every query and methods if the column name change later on.

eric-chau:
Completely out-of-date now, I close it.

PR #366, opened 2015-05-12
[RFR] - Added security rights on GroupController

mickaelandrieu:
* Add security annotations on GroupController
* Rest tests are now executed as sudoer
* Rest tests can now be executed with this command ``phpunit --group=Rest``

### DIFF
--- a/Rest/Controller/GroupController.php
+++ b/Rest/Controller/GroupController.php
@@ -44,6 +44,7 @@ class GroupController extends AbstractRestController
 {
     /**
      * Get all records.
+     * @Rest\Security("is_fully_authenticated() & has_role('ROLE_API_USER')")
      */
     public function getCollectionAction(Request $request)
     {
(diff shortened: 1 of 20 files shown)


Rest/Controller/GroupController.php
```diff
@@ -109,6 +113,7 @@ public function putAction(Group $group, Request $request)
      *  @Assert\NotBlank(message="Name is required"),
      *  @Assert\Length(max=50, minMessage="Maximum length of name is 50 characters")
      * })
+     * @Rest\Security("is_fully_authenticated() & has_role('ROLE_API_USER')")
```
eric-chau: you should test if user can create a group no?

eric-chau:
:+1:, thanks for your contribution @mickaelandrieu

Rest/Controller/GroupController.php
mickaelandrieu: You're right, I will update the expression.

mickaelandrieu:
thank you for review @eric-chau

PR #374, opened 2015-05-14
[Security] Define available roles for BBRoleVoter as container parameter bbapp.securitycontext.bbuser.roles

crouillon:


eric-chau:
@crouillon, I think we should doing it deeper by setting these default roles into ``Security/Authorization/Voter/BBRoleVoter`` cause we are hard using these into ``BackBee\Security\User::getRoles``. What do you think? And we keep the possibility to add more specific role by service container parameters.

mickaelandrieu:
@eric-chau agree

Security/Authorization/Voter/BBRoleVoter.php
```diff
@@ -96,6 +94,6 @@ public function vote(TokenInterface $token, $object, array $attributes)
 
     protected function extractRoles(TokenInterface $token)
     {
-        return $this->_adapter->extractRoles($token);
+        return $token->getRoles();
```
mickaelandrieu: :+1:

Security/Authorization/Voter/BBRoleVoter.php
```diff
@@ -36,22 +36,20 @@
  */
 class BBRoleVoter implements VoterInterface
```
mickaelandrieu: Maybe extends the Symfony's RoleVoter class could be more efficient ?  (https://github.com/symfony/Security/blob/master/Core/Authorization/Voter/RoleVoter.php)

crouillon:
@eric-chau, i think these roles shouldn't be hardly set in User entity. The authentication providers should extends user's persisted roles with :
- ROLE_ACTIVE_USER by BBAuthenticationProvider (that's why i added the public isActiveUser())
- ROLE_API_USER by PublicKeyAuthenticationProvider
That is the reason I pass these roles to the voter by now, and, @mickaelandrieu, yes the voter could be the one of symfony now, it would imply that any role starting with ROLE_ should be supported by it's not so strange
So, if you agree, i'll modify authentication providers to set respective role to user and remove them from the entity and finally change the voter to RoleVoter

crouillon:
One last thing, linked to #376, Security\Authorization\Adapter become unused

eric-chau:
@crouillon, I completely agree with the idea to remove this hard test inside User entity and delegate it to voters. Thank you for taking time to explain it to us :) it's gonna be a great contribution. 
Ok for the unused part I will remove it too.

mickaelandrieu:
@crouillon @eric-chau :+1: agree ! the less code we have to maintain, the more features we can provide to users :)

eric-chau:
@crouillon, please add test for ``SecurityController::authenticateAction``.

eric-chau:
ping @crouillon for #396. Review done with @crouillon, this need to merge to fix authentication issue.

Issue #376, opened 2015-05-14
Remove unused components from project


eric-chau:
- ``Application``
- ``Form``
- unused classes from ``Security`` component
- unused classes from ``Bundle`` component

eric-chau:
I'm on it

mickaelandrieu:
``Importer`` also ?

You can move it outside BackBee core, inside backbee/importer repository maybe ?

eric-chau:
Yeah sure, but i will do separated ticket for these specific task.

eric-chau:
Resolved.

PR #377, opened 2015-05-16
Remove unused components and classes #376

eric-chau:
More details in #376.

Removed unused component from ``backbee/backbee``. Also unused classes are deleted.


mickaelandrieu:
:+1:

Issue #382, opened 2015-05-18
Move Validator component out of project


eric-chau:
``Validator`` component should not be inside of ``backbee/backbee``.

PR #393, opened 2015-05-18
[Validator] moved this component into its own repository #382

eric-chau:
More details on #382. Also removed its dependency in composer.json and its declared services into application's container.


eric-chau:
ping @mickaelandrieu, @fkroockmann.

Btw @fkroockmann, I created a repository into my account but i planned to transfer it into your one cause it's your contribution. Let me know if it's ok for you. What should we do about PR #171? You want to transfer it too?

fkroockmann:
You can merge this PR https://github.com/backbee/BackBee/pull/171, it's not a bug but an improvement. I will do that later. I'm ok with this move :)

PR #398, opened 2015-05-19
[ClassContent] fixed ClassContentRepository::findPagesByContent

eric-chau:
Currently ``ClassContent/Repository/ClassContentRepository::findPagesByContent`` does not return the expected result cause it relied on Doctrine lazy load which did not do the right result. So I resolved it by doing native SQL query.


eric-chau:
Review it please @mickaelandrieu.

mickaelandrieu:
:+1:

Issue #310, opened 2015-04-02
Move importer outside BackBee (and probably deprecate it ?)


mickaelandrieu:
Importer component shouldn't be part of Core as it is only used when we need to import data from legacy "non BackBee" applications into BackBee.

This is very specific actions, and this may be a helper as this is not very useful outside of BackBee CMS scope.

Also, the quality of this part is very bad and this need a big refactoring.

PR #401, opened 2015-05-20
[Importer] removed Importer component from project #310

eric-chau:
More details in #310.

Component new repository: https://github.com/backbee/Importer


mickaelandrieu:
Perfect ! :+1:  (and also :+1:  to deprecate this part of CMS, we better should provide a bundle/library with some interfaces to help developper to create their own importers imo)

eric-chau:
Thanks @mickaelandrieu, I think this PR can really be merged without any problem cause it was really isolated from the rest.

I tagged the backbee/importer package as deprecated. I think it's enough for now, we may provide some interfaces later on to help developers.

PR #413, opened 2015-05-21
fixing mediafolder rest getAction

hbaptiste:
fixing mediafolder rest getAction


eric-chau:
:+1:

mickaelandrieu:
this won't pass the tests, did you have tried on your own machine @hbaptiste ?

eric-chau:
I think I know why it does not work. I will check it for you @hbaptiste

eric-chau:
Oh dude, I didn't notice that you use ``::toArray`` inside ``MediaFolder::jsonSerialize`` method... we have to change it. I will do it with you @hbaptiste.

mickaelandrieu:
Then i don't see why we need another ``jsonSerialize`` method as we already have this helpers in RestController, can you explain ?

eric-chau:
This PR is replaced by PR #417.

Issue #394, opened 2015-05-19
[Tests] autoloader and TestCase must be updated


eric-chau:
They are out of date and contain many bad practices.

PR #414, opened 2015-05-21
[ClassContent] fixed every processes about classcontent and its revision persistences

eric-chau:
This PR aims to fix every issues users met about ``ClassContent`` and ``Revision`` persistence and deletion.


eric-chau:
Currently ``BackBee\Tests\TestCase`` and ``bootstrap.php.dist`` are bad designed so tests does not pass. Tests will be added when the issue #394 will be resolved.

PR #417, opened 2015-05-22
[NestedNode] normalized MediaFolder::jsonSerialize and updated MediaFolderController::getAction 

eric-chau:
normalized ``BackBee\NestedNode\MediaFolder::jsonSerialize`` and updated ``BackBee\Rest\Controller\MediaFolderController::getAction`` to return a valid json response.

Also cleaned every classes which contains method ``::toArray``, ``::serialize`` and ``::unserialize``, they were used by JsonRPC Server and are no longer used now.

ping @hbaptiste you need to adapt MediaFolder in Javascript side.


mickaelandrieu:
tests pass => :+1:

PR #418, opened 2015-05-22
Return code status and not bad strings

mickaelandrieu:
This is an API improvement.

Now Pages return an array of page statuses and not a string, we need to merge BbCoreJS pull request https://github.com/backbee/BbCoreJs/pull/767

### DIFF
--- a/NestedNode/Page.php
+++ b/NestedNode/Page.php
@@ -1402,24 +1402,26 @@ class Page extends AbstractNestedNode implements RenderableInterface, DomainObje
 
     /**
      * @Serializer\VirtualProperty
-     * @Serializer\SerializedName("state_label")
+     * @Serializer\SerializedName("states")
      */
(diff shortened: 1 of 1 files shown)


eric-chau:
I coded this behavior about returning a string instead of an array but if an array suits better with your need go ahead. Also I think an array is more exploitable than an string in this case. 

One last thing, why do you return ``Page`` state integer instead of its label? I'm not sure that Javascript side should be able to interpret what's Php logic in the end. ping @crouillon for this point.

Otherwise :+1: thanks @mickaelandrieu.

mickaelandrieu:
@eric-chau I have also did the pull request in Javascript core client ;)

mickaelandrieu:
backbee/BbCoreJs#767 is merged, and we need this one to be merged too.

Thank you for review @eric-chau

PR #425, opened 2015-05-25
Upgrade group entity to add description

ndufreche:
see backbee/BbCoreJs#765

### DIFF
--- a/Security/Group.php
+++ b/Security/Group.php
@@ -67,6 +67,16 @@ class Group implements DomainObjectInterface
     protected $_name;
 
     /**
+     * Group description.
+     *
+     * @var string
+     * @ORM\Column(type="string", name="description", nullable=true)
+     *
+     * @Serializer\Expose
+     */
+    protected $_description;
+
+    /**
      * @var \Doctrine\Common\Collections\ArrayCollection
      *
      * @Serializer\Expose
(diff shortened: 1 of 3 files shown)


mickaelandrieu:
tests broken :) please fix I need this ``removeUser`` too :p

ndufreche:
Is why it's in WIP :p

Security/Group.php
```diff
@@ -190,6 +224,34 @@ public function addUser(User $user)
...
+    }
+
+    /**
+     * @codeCoverageIgnore
+     *
+     * @param \BackBee\Security\User $user
+     *
+     * @return \BackBee\Security\Group
+     */
+    // public function removeUser(User $user)
```
ndufreche: I don't know can I had of this historical function. The good function to remove an entity from an Many To Many in doctrine is with "removeElement" function

mickaelandrieu: I agree with you, it's ``removeElement()`` function.

ndufreche: http://doctrine-orm.readthedocs.org/en/latest/reference/working-with-associations.html#removing-associations from the doc :D

ndufreche: Before an research in the Core php I don't found a usage of this function, I remove it.

mickaelandrieu:
@ndufreche this need a squash :)

ndufreche:
Just I watch the Travis validation.then(squash, resolve)

mickaelandrieu:
:+1:

eric-chau:
Review done with @ndufreche which agrees with the fact that we can improve the logic so I created an issue #428. We decide to merge it like this at moment cause it will imply too much change in PHP and JS sides. :+1:

PR #427, opened 2015-05-25
[WIP][REST] User Controller patch test

ndufreche:
Is a split off my previous PR #425 because I have a problem on my test.

(Only the last commit is important)

My current error:
```
Use of undefined constant self - assumed 'self'
```


eric-chau:
This PR won't be update anymore. I close it.

PR #429, opened 2015-05-26
[NestedNode] + Fix hasChildren function into Page

fkroockmann:
Currenlty, hasChildren don't count a deleted page.

### DIFF
--- a/NestedNode/Page.php
+++ b/NestedNode/Page.php
@@ -1530,8 +1530,18 @@ class Page extends AbstractNestedNode implements RenderableInterface, DomainObje
      * @Serializer\VirtualProperty
      * @Serializer\Type("boolean")
      */
-    public function hasChildren()
+    public function hasChildren($ignoreDeleted = true)
     {
+        if ($ignoreDeleted) {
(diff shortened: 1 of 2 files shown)


NestedNode/Page.php
```diff
@@ -1532,7 +1532,13 @@ private function createNewDefaultContent($classname, $mainzone = false)
      */
     public function hasChildren()
```
eric-chau: we should change the method signature to ``public function hasChildren($ignoreDeleted = true)``? So we can also check for deleted children if needed.

eric-chau:
Change ``BackBee\NestedNode\Page::hasChildren`` method signature please and add some tests to this and we are good to go with it =) :+1:

NestedNode/Tests/PageTest.php
```diff
@@ -345,6 +345,74 @@ public function testSetLayout()
     }
 
     /**
+     * @covers BackBee\NestedNode\Page::hasChildren
```
eric-chau: remove this annotation please =)

eric-chau:
Btw technically it's not a fix but an improvement =)

PR #431, opened 2015-05-27
[Rest] updated ClassContentController::findContentsByCriterias to also hydrate contents draft

eric-chau:
Currently content's draft is not hydrated when we get content collection if we are logged, this PR fixes it.


mickaelandrieu:
:+1:

PR #432, opened 2015-05-27
Added new authorisations on Page to entity patcher

mickaelandrieu:
* added authorisations of PATCH on archiving/publishing

REFS https://github.com/backbee/BbCoreJs/issues/746


eric-chau:
:+1: nice catch, we may be able to break the dependence to JMS serializer.

PR #439, opened 2015-05-28
[Resources] cleaned every unused files and updated error layouts

eric-chau:
- removed outdated images
- removed outdated layouts
- updated and translated error layouts
- added favicon to resources


mickaelandrieu:
:+1:

PR #440, opened 2015-05-28
Throw InsufficientAuthenticationException rather than AccessDeniedException on granted failed

crouillon:
Also send 403 status code on InsufficientAuthenticationException throwing
+ Fix user right on PageController


eric-chau:
:+1: nice improvements

PR #443, opened 2015-05-28
Switch AccessDenied to InsufficientAuthentication in  user controler

ndufreche:
To avoid disconnection in toolbar bundle if you are not allowed to view a user


ndufreche:
screenshot from 2015-05-28 17 23 24

mickaelandrieu:
OK, thank you @ndufreche

Issue #308, opened 2015-04-02
Update our dependencies: Symfony 2.6 and Doctrine ORM 2.5


mickaelandrieu:
Hi,

the update of Symfony components we use should be easy, and I think the tip provided by @javiereguiluz (previously @rande has revealed on twitter) => http://blog.insight.sensiolabs.com/2014/12/22/making-symfony-bootable-with-dbal-2-5.html should help us to migrate an so on we can apply the 2nd level cache and improve our performances :)

mickaelandrieu:
It's now very important as Symfony 2.5 won't be maintained at the end of June.

We need to update from Symfony 2.7 which is the new LTS (maintained until 2019)

PR #448, opened 2015-06-03
[RFR] Updated to Symfony 2.7

mickaelandrieu:
(REFS  #308)

Steps before a complete migration to Symfony 2.7 components, an exhaustive tasks list is available here: https://gist.github.com/mickaelandrieu/5211d0047e7a6fbff925

This is now the complete update to Symfony 2.7.

Thanks to the Symfony PHPUnit brige, we can delay the complete upgrade of our code and only care about real software errors (the complete upgrade should be planned, in order to be compliant with Symfony3).

For now, there is only some errors to fix, in order to be able to use the new Symfony 2.7 LTS.


mickaelandrieu:
@eric-chau I'm not sure but I think this update is a BC on PageListener.

Validator implements ValidatorInterface, so if we use ValidatorInterface the listener can use ValidatorRecursive object, and this is this object which is used in Symfony 2.7

The behavior is the same (no tests updated), and I don't think this listener is part of public api... what do you think ?

mickaelandrieu:
@eric-chau @crouillon remains a lot of deprecations to be compliant with Symfony 3 components, but we have time (almost 2,5 y) to reduce this technical debt.

So I suggest to accept this pull request and fix our "Symfony" dependencies for BackBee 1.x .

crouillon:
:+1:

PR #451, opened 2015-06-08
[layout] Avoid infinite loop when picpath is egal to null

ndufreche:
When a layout picpath is null we try to unlink a folder that action create an infinite loop because the while condition stay every time true :

```php
true === file_exists($thumbnailfile) && true === is_writable($thumbnailfile)
```

To avoid this problem I have use `is_file` because it's more restrictive than `file_exists`.

### DIFF
--- a/Site/Repository/LayoutRepository.php
+++ b/Site/Repository/LayoutRepository.php
@@ -210,9 +210,12 @@ class LayoutRepository extends EntityRepository
     public function removeThumbnail(Layout $layout, BBApplication $app)
     {
         $thumbnailfile = $layout->getPicPath();
+        if (empty($thumbnail)) {
+            return true;
+        }
         File::resolveFilepath($thumbnailfile, null, array('include_path' => $app->getResourceDir()));
 
-        while (true === file_exists($thumbnailfile) && true === is_writable($thumbnailfile)) {
+        while (true === is_file($thumbnailfile) && true === is_writable($thumbnailfile)) {
             @unlink($thumbnailfile);
 
             $thumbnailfile = $layout->getPicPath();


Site/Repository/LayoutRepository.php
```diff
@@ -212,7 +212,7 @@ public function removeThumbnail(Layout $layout, BBApplication $app)
         $thumbnailfile = $layout->getPicPath();
         File::resolveFilepath($thumbnailfile, null, array('include_path' => $app->getResourceDir()));
 
-        while (true === file_exists($thumbnailfile) && true === is_writable($thumbnailfile)) {
+        while (true === is_file($thumbnailfile) && true === is_writable($thumbnailfile)) {
```
eric-chau: We don't need to test php internal function returned value type: ``while (is_file($thumbnailfile) && is_writable($thumbnailfile)) {`` is enough.
And this change is not enough good to avoid infinite loop.

Site/Repository/LayoutRepository.php
```diff
@@ -212,7 +212,7 @@ public function removeThumbnail(Layout $layout, BBApplication $app)
...
             @unlink($thumbnailfile);
```
eric-chau: To avoid infinite loop, do something like this:

```php
if (false === unlink($thumbnailfile)) {
    throw new \Exception(sprintf('An error occured while we try to delete the file `%s`.', $thumbnailfile));
}
```

I don't know if we should use ``RuntimeException`` or ``LogicException``?

mickaelandrieu: LogicException come from user, RuntimeException come from server => RuntimeException

ndufreche: No because picpath is not required value, then we can't throw an exception because this value is not set. Them the side effect of add an acception : I can't delete a Layout because I don't set an non required value, this is not friendly.

Site/Repository/LayoutRepository.php
ndufreche: Yes is not the best but it work fine.

Site/Repository/LayoutRepository.php
eric-chau: The problem is not at the moment we call for ``unlink`` cause it's already too late. We just need to add one more test:

```php
public function removeThumbnail(Layout $layout, BBApplication $app)
{
     $thumbnailfile = $layout->getPicPath();
     if (empty($thumbnailfile)) {
        return false; // or true? What the meaning of returning a boolean after all? Since the thumbnail is not required
    }

    // ...
```

ndufreche: Ok I can do that, but i think is good idea to keep is_file test in case of picpath is badly set and do a reference to a folder (it's really extreme but stay possible).

ndufreche: Updated

mickaelandrieu:
Thank you @ndufreche

PR #453, opened 2015-06-10
Fix getMode function in AbstractClassContent

fkroockmann:
Currenlty, this function use a old mangement of parameters

### DIFF
--- a/ClassContent/AbstractClassContent.php
+++ b/ClassContent/AbstractClassContent.php
@@ -882,22 +882,8 @@ abstract class AbstractClassContent extends AbstractContent
      */
     public function getMode()
     {
-        $rendermode = null;
-
-        if (is_array($this->getParam('rendermode'))) {
-            $rendermode = (array) $this->getParam('rendermode');
-            $rendermode = array_pop($rendermode);
-
-            if (isset($rendermode['rendertype'])) {
-                switch ($rendermode['rendertype']) {
-                    case 'select':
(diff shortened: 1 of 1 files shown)


ClassContent/AbstractClassContent.php
```diff
@@ -882,22 +882,8 @@ public function acceptSubcontent($var)
...
-                    case 'select':
-                        $rendermode = $rendermode['selected'];
-                        break;
-                }
-            }
-        }
-
-        return $rendermode;
+        $rendermode = $this->getParamValue('rendermode');
+        return (is_array($rendermode)) ? reset($rendermode) : null;
```
eric-chau: Follow our coding style:

```php
$rendermode = $this->getParamValue('rendermode');

return is_array($rendermode) ? reset($rendermode) : null;
```

eric-chau:
Nice catch, thank you @fkroockmann :+1:

ClassContent/AbstractClassContent.php
fkroockmann: It's necessary cause the reset function use this param as reference.

eric-chau: You misunderstood what I said cause I was talking about the coding style but whatever, it's already merged.

eric-chau:
Btw could be really nice to add some tests so we will never get regression about it anymore.

PR #454, opened 2015-06-11
[WIP][ClassContent] updated CategoryManager and content rest controller to accept 'only_visible' option

eric-chau:
With this ``only_visible`` option we can easily display into content selector only the contents that are visibles and availables as drag and droppable content.

@hbaptiste this will be useful for you, you will only need to add a query parameter into the url, example: ``GET /rest/1/classcontent?category=Article&only_online``

This PR still in progress, I will add some tests.

### DIFF
--- a/Rest/Controller/ClassContentController.php
+++ b/Rest/Controller/ClassContentController.php
@@ -104,7 +104,12 @@ class ClassContentController extends AbstractRestController
             $start = 0;
         } else {
             if (null !== $categoryName) {
-                $contents = $this->getClassContentByCategory($categoryName, $start, $count);
+                $contents = $this->getClassContentByCategory(
+                    $categoryName,
+                    $request->request->has('only_visible'),
+                    $start,
+                    $count
+                );
             } else {
                 $classnames = $this->getClassContentManager()->getAllClassContentClassnames();
                 $contents = $this->findContentsByCriterias($classnames, $start, $count);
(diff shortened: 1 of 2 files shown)


Rest/Controller/ClassContentController.php
```diff
@@ -565,9 +570,13 @@ private function getClassContentByTypeAndUid($type, $uid, $hydrateDraft = false,
      *
      * @return null|Paginator
      */
-    private function getClassContentByCategory($name, $start, $count)
+    private function getClassContentByCategory($name, $onlyVisible, $start, $count)
```
fkroockmann: You break the compatibility no ?

mickaelandrieu: yes, and no.

This is a private function so this is not part of "public " api.

But, I suggest this function definition => 
``
private function getClassContentByCategory($name, $start, $count, $onlyVisible)
``

eric-chau: This is not a BC cause it's a private method and I let this as second argument for argument order coherence.

mickaelandrieu:
@eric-chau tell me when you consider this PR as done.

mickaelandrieu:
ping @eric-chau let's take time to finish this one tommorrow :)

eric-chau:
This PR was developed for @hbaptiste needs but seems that it's not the case anymore. I close this PR.

PR #455, opened 2015-06-11
Dispatch rest event onKernelController and onKernelResponse

fkroockmann:
If the controller extends AbstractRestController, an event is dispatched

For Request: 'rest.controllername.actionname.request'
For Response: 'rest.controllername.actionname.response'

### DIFF
--- a/Config/events.yml
+++ b/Config/events.yml
@@ -90,6 +90,7 @@ kernel.exception:
 
 kernel.controller:
     listeners:
+        - [@rest.listener.event_dispatcher, onKernelController]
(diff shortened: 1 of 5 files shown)


fkroockmann:
Wait before merge plz

fkroockmann:
Ok it's already done in Frontcontroller::invokeAction.... :(

mickaelandrieu:
NOOOOB

eric-chau:
Yes I ported this feature to ``FrontController``. Previously it was only done for JsonRPC request.

PR #458, opened 2015-06-15
Config component review

mickaelandrieu:
Some refactorings since I have read this component to write docs

### DIFF
--- a/Config/Config.php
+++ b/Config/Config.php
@@ -60,7 +60,7 @@ class Config implements DispatchTagEventInterface, DumpableServiceInterface
     const CONFIG_FILE = 'config';
 
     /**
-     * System extention config file.
+     * System extension config file.
      *
      * @var string
      */
@@ -141,9 +141,11 @@ class Config implements DispatchTagEventInterface, DumpableServiceInterface
     /**
      * Class constructor.
      *
-     * @param string                                 $basedir   The base directory in which look for config files
-     * @param \BackBee\Cache\AbstractCache           $cache     Optional cache system
-     * @param \BackBee\DependencyInjection\Container $container
+     * @param string                                 $basedir       The base directory in which look for config files
+     * @param \BackBee\Cache\AbstractCache           $cache         Optional cache system
+     * @param \BackBee\DependencyInjection\Container $container     The BackBee Container
+     * @param boolean                                $debug         The debug mode
+     * @param array                                  $yml_to_ignore List of yaml filename to ignore form loading/parsing process
      */
     public function __construct($basedir, AbstractCache $cache = null, Container $container = null, $debug = false, array $yml_to_ignore = array())
     {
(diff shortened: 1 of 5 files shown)


Config/Config.php
```diff
@@ -404,9 +408,9 @@ public function extend($basedir = null, $overwrite = false)
         }
 
         if (
-            false === empty($this->environment)
-            && false === strpos($this->environment, $basedir)
-            && true === file_exists($basedir.DIRECTORY_SEPARATOR.$this->environment)
+            !empty($this->environment)
+            && !strpos($this->environment, $basedir)
```
eric-chau: You cannot change it this way cause strpos can return ``0``or ``false`` which doesn't mean the same thing. You don't need to change this line.

Config/Config.php
```diff
@@ -577,7 +580,7 @@ private function getYmlFiles($basedir)
 
         $default_file = $basedir.DIRECTORY_SEPARATOR.self::CONFIG_FILE.'.'.self::EXTENSION;
```
eric-chau: Change the snake case please =)

Config/Config.php
mickaelandrieu: damn PHP API :/ I will revert this change.

Config/Config.php
```diff
@@ -598,11 +601,13 @@ private function getYmlFiles($basedir)
      *
      * @param string $basedir The base directory
      *
-     * @throws \BackBee\Config\Exception\InvalidBaseDirException Occurs if the base directory cannont be read
+     * @throws \BackBee\Config\Exception\InvalidBaseDirException Occurs if the base directory can't be read
      */
     private function loadFromBaseDir($basedir, $overwrite = false)
     {
-        foreach ($this->getYmlFiles($basedir) as $filename) {
+        $files = $this->getYmlFiles($basedir);
```
eric-chau: why did you to declare a variable instead of using the result directly into the ``foreach`` loop?

Config/Config.php
eric-chau: In this case it's not about PHP =) it's kinda logical to me ^^

Config/Config.php
mickaelandrieu: I dont want to make a function call each time, I'm sure Insight will complain :cat:

eric-chau: But it calls the function only once ^^ the ``foreach`` loop is performed only when the result is computed.

eric-chau:
I don't know if your work is over about this PR but I already did a first review pass. Ping me if you need another one later.

Config/Config.php
mickaelandrieu: Yup, strpos is misused but we dont have specific (boolean return) function for this use case in PHP.

I mean, a kind of ``exists`` which always return a boolean is missing ^^

eric-chau:
:+1:

mickaelandrieu:
@eric-chau I have done your improvements, feel free to merge it if it's OK :)

mickaelandrieu:
@eric-chau and this one, Ok for merge :cat: ?

eric-chau:
@mickaelandrieu, yes nice improvements, thank you.

mickaelandrieu:
thanks for review :)

PR #459, opened 2015-06-15
[WIP]Creating bundleAdminUrl helper

ndufreche:
This is an helper to generate easily bundle rest URL.

```twig
{{ this.bundleAdminUrl('demo.admin.index', {'id'},  {'optional': 'query'})|row }}
```

have to render `http://backbee.url/rest/1/bundle/demo/admin/index/id?optional=query`

To help the implementation when backbee is in debug mode we check all parameters if they are goodly defined


Renderer/Helper/bundleAdminUrl.php
```diff
@@ -0,0 +1,103 @@
...
+{
+
+    /**
+     * @param  [string] $route      [route is composed by the bundle, controller and action name separated by a dot]
+     * @param  array    $parameters [optional parameters]
+     * @param  array    $query      [optional query parameters]
+     *
+     * @return [string]             [url]
+     */
+    public function __invoke($route, $parameters = [], $query = [])
```
eric-chau: Force the type of ``$parameters`` to ``array`` so you don't need to check ``if (count($parameters) !== 0) {`` and you can call foreach directly.

Renderer/Helper/bundleAdminUrl.php
```diff
@@ -0,0 +1,103 @@
...
+
+        $url = 'rest/1/bundle/'.$bundle.'/'.$controller.'/'.$action;
+
+        if (count($parameters) !== 0) {
+            foreach ($parameters as $parameter) {
+                $url = $url.'/'.$parameter;
+            }
+        }
+
+        if (count($query) !== 0) {
```
eric-chau: Same comment for ``$query``, force it to be type of ``array``.

Renderer/Helper/bundleAdminUrl.php
```diff
@@ -0,0 +1,103 @@
...
+            $i = 0;
+            $url = $url.'?';
+            foreach ($query as $key => $value) {
```
eric-chau: You don't need to do a ``foreach``, an implode is good enough and the syntax is lightweight:

``$url = $url.'?'.implode('&', $query);``

Renderer/Helper/bundleAdminUrl.php
```diff
@@ -0,0 +1,103 @@
...
+     * @param  [string] $controller [controller name]
+     * @param  [string] $action     [action name]
+     *
+     * @throws [Exception] [Bad configuration]
+     */
+    private function checkParameters($bundle, $controller, $action)
+    {
+        $app = $this->_renderer->getApplication();
+
+        if (!$app->getContainer()->has('bundle.'.$bundle)) {
```
eric-chau: You can use ``BackBee\Bundle\BundleInterface::BUNDLE_SERVICE_ID_PATTERN`` to compute the bundle service identifier: 
```php
if (!$app->getContainer()->has(str_replace('%bundle_name%', strtolower($id), BundleInterface::BUNDLE_SERVICE_ID_PATTERN))) {
    // ...
}
```
Doing it this way will avoid you BC if later on we decide to change the bundle service identifier pattern.

Renderer/Helper/bundleAdminUrl.php
```diff
@@ -0,0 +1,103 @@
...
+            throw new \Exception($bundle.' doesn\'t exists', 1);
+        }
+
+        $config = $app->getContainer()->get('bundle.'.$bundle.'.config');
```
eric-chau: Same for bundle's config service identifier, there is a constant too: ``BundleInterface::CONFIG_SERVICE_ID_PATTERN``

Renderer/Helper/bundleAdminUrl.php
```diff
@@ -0,0 +1,103 @@
...
+
+        if (!array_key_exists('controller', $config)) {
```
eric-chau: Be used to use ``isset`` instead of ``array_key_exists`` cause this last one perform a function call while ``isset`` does not. Use ``array_key_exists`` only in the cases you matter about ``null`` value too.

Renderer/Helper/bundleAdminUrl.php
```diff
@@ -0,0 +1,103 @@
...
+            throw new \Exception('No controller definition in '.$bundle.' bundle configuration');
+        }
+
+        if (!array_key_exists($controller, $config['controller'])) {
+            throw new \Exception($controller.' controller is undefinned in '.$bundle.' bundle configuration');
```
eric-chau: ``Undefinned`` => ``Undefined``

Renderer/Helper/bundleAdminUrl.php
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        $controller = new {'\\'.$config['controller']['$controller']}();
```
eric-chau: I need to write documentation about this part but you should not doing the test here by yourself but using ``AbstractBundle::getExposedActionsMapping`` instead.

Renderer/Helper/bundleAdminUrl.php
ndufreche: Ok.
But what is the impact on the bundle name passed to the function.
`demo` will be transform into `demobundle` and that depends off the name you give to your bundle into bundle.yml .
Is right ?

eric-chau: Don't worry ``demo`` will become ``bundle.demo``. Here the full declaration of the constant in ``BundleInterface``: ``const BUNDLE_SERVICE_ID_PATTERN = 'bundle.%bundle_name%';``

Renderer/Helper/bundleAdminUrl.php
```diff
@@ -0,0 +1,95 @@
...
+     * @param  array    $query      [optional query parameters]
+     *
+     * @return [string]             [url]
+     */
+    public function __invoke($route, Array $parameters = [], Array $query = [])
+    {
+        list($bundle, $controller, $action) = split('.', $route);
+
+        if ($this->_renderer->getApplication()->isDebugMode()) {
+            $this->checkParameters($bundle, $controller, $action)
```
mickaelandrieu: are you kidding ? ``;`` missing.

Renderer/Helper/bundleAdminUrl.php
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        $url = 'rest/1/bundle/'.$bundle.'/'.$controller.'/'.$action;
```
mickaelandrieu: Hum I'm not easy with this kind of stuff, if we publish a version 2 of API or change the url this will break without any information.

Renderer/Helper/bundleAdminUrl.php
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        if (!isset($config['controller'][$controller])) {
+            throw new \Exception($controller.' controller is undefinned in '.$bundle.' bundle configuration');
+        }
+
+        $controller = new {'\\'.$config['controller']['$controller']}();
+
+        if (!method_exists($controller, $action.'Action')) {
+            throw new \Exception$controller.' doesn\'t exists', 1);
```
mickaelandrieu: ``(`` missing, I don't understand why you pass "1" as a second argument of this Exception.

Renderer/Helper/bundleAdminUrl.php
```diff
@@ -0,0 +1,95 @@
...
+     * Check if each parameters are correctly setted
+     * @param  [string] $bundle     [bundle name]
+     * @param  [string] $controller [controller name]
+     * @param  [string] $action     [action name]
+     *
+     * @throws [Exception] [Bad configuration]
+     */
+    private function checkParameters($bundle, $controller, $action)
+    {
+        $app = $this->_renderer->getApplication();
```
mickaelandrieu: Declare a ``$container`` instead.

```php
$container = $this->_renderer->getApplication()->getContainer();

if(!$container->has(...))
```

mickaelandrieu:
Also, this could be usefull to complete and reuse the Bundle exceptions (https://github.com/backbee/BackBee/blob/master/Bundle/Exception/BundleException.php), what do you think ?

Renderer/Helper/bundleAdminUrl.php
```diff
@@ -0,0 +1,95 @@
...
+ */
+class bundleAdminUrl extends AbstractHelper
+{
+
+    /**
+     * @param  [string] $route      [route is composed by the bundle, controller and action name separated by a dot]
+     * @param  array    $parameters [optional parameters]
+     * @param  array    $query      [optional query parameters]
+     *
+     * @return [string]             [url]
```
mickaelandrieu: /c @eric-chau I'm not sure, is it valid phpdoc ?

PR #464, opened 2015-06-17
Fix annotation autoloading which is broken by Symfony 2.7

fkroockmann:
Fix annotation autoloading brokes by Symfony 2.7

### DIFF
--- a/BBApplication.php
+++ b/BBApplication.php
@@ -1061,12 +1061,15 @@ class BBApplication implements ApplicationInterface, DumpableServiceInterface, D
 
         // annotations require custom autoloading
         AnnotationRegistry::registerAutoloadNamespaces([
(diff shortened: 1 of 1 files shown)


eric-chau:
This issue appeared with the update of Symfony dependencies to 2.7. Symfony updated the folder's structure of the validator component so it changed the way we have to autoload validator's annotations. :+1:

BBApplication.php
```diff
@@ -1070,6 +1069,10 @@ private function initAnnotationReader()
                 return class_exists($classname);
             }
 
+            if (0 === strpos($classname, 'Symfony\Component\Validator\Constraints')) {
+                return class_exists($classname);
```
mickaelandrieu: this is weird, why do we need to do this kind of stuff ?

PR #466, opened 2015-06-18
[RFR]Creating bundleAdminUrl helper

ndufreche:
This is an helper to generate easily bundle rest URL.

```twig
{{ this.bundleAdminUrl('demo.admin.index', {'id'},  {'optional': 'query'})|row }}
```

have to render `http://backbee.url/rest/1/bundle/demo/admin/index/id?optional=query`

To help the implementation when backbee is in debug mode we check all parameters if they are goodly defined

### DIFF
--- a/Bundle/BundleControllerResolver.php
+++ b/Bundle/BundleControllerResolver.php
@@ -1,0 +1,75 @@
+<?php
+
+/*
+ * Copyright (c) 2011-2015 Lp digital system
+ *
+ * This file is part of BackBee.
+ *
+ * BackBee is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * BackBee is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with BackBee. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * @author Charles Rouillon <charles.rouillon@lp-digital.fr>
+ */
+
+namespace BackBee\Bundle;
+
+use BackBee\Bundle\Exception\BundleConfigurationException;
+use BackBee\DependencyInjection\ContainerInterface;
+
+/**
+ * @category    BackBee
+ *
+ * @author      Nicolas Dufreche <nicolas.dufreche@lp-digital.fr>
+ */
+class BundleControllerResolver
+{
+    private $container;
+
+    public function __construct(ContainerInterface $application)
+    {
+        $this->container = $container;
+    }
+
+    private function computeBundleName($name)
+    {
+        return str_replace('%bundle_name%', strtolower($name), BundleInterface::BUNDLE_SERVICE_ID_PATTERN);
+    }
+
+    /**
+     * Check if each parameters are correctly setted
+     *
+     * @param  string $bundle       bundle name
+     * @param  string $controller   controller name
+     * @param  string $action       action name
+     *
+     * @throws Exception            Bad configuration
+     */
+    private function resolve($bundle, $controller)
+    {
+        if (!$this->container->has($this->computeBundleName($bundle))) {
+            throw new BundleConfigurationException($bundle.' doesn\'t exists', BundleConfigurationException::BUNDLE_UNDLECLARED);
+        }
+
+        $config = $this->container->get($this->computeBundleName($bundle));
+
+        if (!isset($config['controller'])) {
+            throw new BundleConfigurationException('No controller definition in '.$bundle.' bundle configuration', BundleConfigurationException::CONTROLLER_SECTION_MISSING);
+        }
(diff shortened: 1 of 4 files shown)


Renderer/Helper/bundleAdminUrl.php
```diff
@@ -0,0 +1,94 @@
...
+     */
+    public function __invoke($route, Array $parameters = [], Array $query = [])
+    {
+        list($bundle, $controller, $action) = split('.', $route);
+
+        if ($this->_renderer->getApplication()->isDebugMode()) {
+            $this->checkParameters($bundle, $controller, $action);
+        }
+
+        $url = 'rest/1/bundle/'.$bundle.'/'.$controller.'/'.$action;
```
ndufreche: > @mickaelandrieu
Hum I'm not easy with this kind of stuff, if we publish a version 2 of API or change the url this will break without any information.

ndufreche:
> @mickaelandrieu 
Also, this could be usefull to complete and reuse the Bundle exceptions (https://github.com/backbee/BackBee/blob/master/Bundle/Exception/BundleException.php), what do you think ?

Renderer/Helper/bundleAdminUrl.php
```diff
@@ -0,0 +1,94 @@
...
+ */
+class bundleAdminUrl extends AbstractHelper
+{
+
+    /**
+     * @param  [string] $route      [route is composed by the bundle, controller and action name separated by a dot]
+     * @param  array    $parameters [optional parameters]
+     * @param  array    $query      [optional query parameters]
+     *
+     * @return [string]             [url]
```
ndufreche: > @mickaelandrieu 
/c @eric-chau I'm not sure, is it valid phpdoc ?

eric-chau: yeah @ndufreche can you remove the square bracket please? (``[]``)

ndufreche: yes

ndufreche:
Yes using global Exception is not a good idea.
Them I have a question, I use BundleException or I create a new Exception like BundleConfigurationException ?

mickaelandrieu:
Ok for a new BundleConfigurationException :+1:

ndufreche:
Exception updated

Renderer/Helper/bundleAdminUrl.php
mickaelandrieu: @ndufreche can you pass a fourth argument called (for instance) ``$endpoint`` with default value ``rest/1/bundle``.

This way we are probably supporting a future v2 of REST api.

mickaelandrieu:
@eric-chau any idea on "how" to test it ? This will become a very important part of new Bundle application, we need to be sure this work perfectly on all scenarios.

Renderer/Helper/bundleAdminUrl.php
ndufreche: There no service for that or an config parameter ?
I think the fourth argument is not a good idea too much constraining for the developer.

mickaelandrieu: Agree to all implementations which remove this "raw" value inside the class :)

eric-chau: In application container you have this parameter ``bbapp.rest_api.path`` which currently return ``/rest/``. Maybe we should introduce a parameter into container about the version of the API.

ndufreche:
@mickaelandrieu I think, I can tested in the DemoBundle context. Or simply by doing some rendering test.

Renderer/Helper/bundleAdminUrl.php
ndufreche: Ok thanks @eric-chau

mickaelandrieu:
@ndufreche in backbee/BackBee, you can also test bundles. Take a look to Rest\BundleController tests :)

Renderer/Helper/bundleAdminUrl.php
```diff
@@ -0,0 +1,84 @@
...
+    {
+        list($bundle, $controller, $action) = split('.', $route);
+
+        if ($this->_renderer->getApplication()->isDebugMode()) {
+            $this->checkParameters($bundle, $controller, $action);
+        }
+
+        $container = $this->_renderer->getApplication()->getContainer();
+
+        $url = $container->get('bbapp.rest_api.path').$container->get('bbapp.rest_api.version').'/bundle/'.$bundle.'/'.$controller.'/'.$action;
```
ndufreche: @mickaelandrieu is better ?

mickaelandrieu: @ndufreche yes ! I'm a little bit afraid of ``/`` but I think PHP can handle it, if we need to improve this later this wont change anything.

A little step for you, a valuable improvement for the future of BackBee :bear:

Bundle/BundleControllerResolver.php
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * @category    BackBee
+ *
+ * @author      Nicolas Dufreche <nicolas.dufreche@lp-digital.fr>
+ */
+class BundleControllerResolver
+{
+    private $container;
+
+    public function __construct(ContainerInterface $application)
```
crouillon: $container not $application I think

ndufreche:
This PR is no more updated, all update are available in => #470
This PR is now closed

Issue #295, opened 2015-03-27
Apply the coding styles of Symfony


mickaelandrieu:
http://symfony.com/doc/current/contributing/code/standards.html

* Interfaces should be suffixed by Interface
* abstract classes should be prefixed by Abstract
* Exception classes should be suffixed by Exception

nicolas-bastien:
@mickaelandrieu want some help on this ?

mickaelandrieu:
@nicolas-bastien yes, maybe a tool can make that for us, what do you think ?

mickaelandrieu:
done.

Issue #53, opened 2014-12-08
Set listener property of Workflow/State from nullable false to true


mickaelandrieu:

mickaelandrieu:
@eric-chau any idea of what we are supposed to do to resolve this one ?

eric-chau:
This issue is resolved by PR #469.

PR #469, opened 2015-06-18
[RFR] [Workflow] updated PageListener and state ListenerInterface

eric-chau:
This PR aims to resolve https://github.com/backbee/backbee-standard/issues/295 to resolve and improves ``BackBee\Workflow\State`` and ``BackBee\Workflow\Listener\PageListener`` by adding tests; it also fix the way we handle state's listener.

### DIFF
--- a/Workflow/Listener/PageListener.php
+++ b/Workflow/Listener/PageListener.php
@@ -24,6 +24,9 @@
 namespace BackBee\Workflow\Listener;
 
 use BackBee\Event\Event;
+use BackBee\NestedNode\Page;
+
+use Doctrine\ORM\Event\PreUpdateEventArgs;
 
 /**
  * Listener to page events.
@@ -33,7 +36,7 @@ use BackBee\Event\Event;
  * @copyright   Lp digital system
  * @author      d.Bensid <djoudi.bensid@lp-digital.fr>
  */
-class pageListener
+class PageListener
 {
     /**
      * Occur on nestednode.page.preupdate events.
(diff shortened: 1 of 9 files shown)


Workflow/Listener/PageListener.php
```diff
@@ -47,41 +50,30 @@ public static function onPreUpdate(Event $event)
...
-        if (is_a($eventArgs, 'Doctrine\ORM\Event\PreUpdateEventArgs')) {
+        if ($eventArgs instanceof PreUpdateEventArgs) {
             if ($eventArgs->hasChangedField('_workflow_state')) {
                 $old = $eventArgs->getOldValue('_workflow_state');
                 $new = $eventArgs->getNewValue('_workflow_state');
 
-                if (null !== $new && null !== $new->getListener()) {
-                    $new->getListener()->arrivedInState($event);
+                if (null !== $new && null !== $listener = $new->getListenerInstance()) {
+                    $listener->switchToState($event);
```
crouillon: :+1:

Workflow/State.php
```diff
@@ -155,11 +160,11 @@ public function getCode()
      */
     public function setCode($code)
     {
-        if (false === Numeric::isInteger($code)) {
+        if (!Numeric::isInteger($code)) {
             throw new InvalidArgumentException('The code of a workflow state has to be an integer');
         }
 
-        $this->_code = $code;
+        $this->_code = (int) $code;
```
mickaelandrieu: Why do you need to cast ? You already have checked that ``$code`` is an integer.

eric-chau: If you look at the code of ``Numeric::isInteger`` it also considers string like ``'123'`` as valid numeric that's why I cast it into integer.

Workflow/State.php
```diff
@@ -198,6 +203,21 @@ public function getListener()
     }
 
     /**
+     * Creates an instance of the listener based on the provided namespace and return it or null
+     * if listener namespace is not setted.
+     *
+     * @return null|ListenerInterface
```
crouillon: ListenerInterface|NULL i think

Workflow/State.php
```diff
@@ -251,7 +288,7 @@ public function setListener($listener = null)
      */
     public function getLayoutUid()
     {
-        return null !== $this->getLayout() ? $this->getLayout()->getUid() : null;
+        return $this->getLayout() ? $this->getLayout()->getUid() : null;
```
mickaelandrieu: What happen if ``$this->_layout`` is null ?

Workflow/State.php
```diff
@@ -229,14 +249,31 @@ public function setLayout(Layout $layout = null)
...
      */
     public function setListener($listener = null)
     {
-        $this->_listener = $listener;
+        if (null !== $listener && !is_object($listener) && !is_string($listener)) {
+            throw new \InvalidArgumentException(sprintf(
+                'Workflow state listener must be type of null, object or string, %s given.',
+                gettype($listener)
+            ));
+        }
```
crouillon: :+1:

Workflow/Tests/PageListenerTest.php
```diff
@@ -0,0 +1,117 @@
+<?php
+
```
mickaelandrieu: Header class missing

Workflow/State.php
```diff
@@ -229,14 +249,31 @@ public function setLayout(Layout $layout = null)
...
+
+        if (null !== $listener && !is_subclass_of($listener, 'BackBee\Workflow\ListenerInterface')) {
```
crouillon: instanceof works also with interfaces, better?

Workflow/State.php
eric-chau: It returns ``null`` cause it's a condition and ``null`` is equivalent to ``false``.

Workflow/Tests/PageListenerTest.php
eric-chau: Thanks =)

Workflow/Listener/PageListener.php
eric-chau: Sounds better I update it right now ;)

Workflow/State.php
```diff
@@ -229,14 +249,31 @@ public function setLayout(Layout $layout = null)
...
+            throw new \LogicException(sprintf(
+                'Workflow state listener must implement %s.',
+                'BackBee\Workflow\ListenerInterface'
+            ));
+        }
+
+        $this->_listener = is_object($listener) ? get_class($listener) : $listener;
+        if (null === $this->_listener) {
+            $this->listenerInstance = null;
```
crouillon: should be reset for each settings of _listener even if not null

Workflow/State.php
mickaelandrieu: o_O I was pretty sure I already have fixed this behavior ... I think we need to pach isInteger instead of patch the code here.

What is the interest of ``isInteger`` if it returns ``true`` for strings ?

Workflow/State.php
eric-chau: You're right, nice catch thank you

Workflow/State.php
eric-chau: I'm not sure we should change ``Numeric::isInteger`` behavior, talk about it with @crouillon first. I change it to use ``is_int()``.

Workflow/State.php
eric-chau: =)

eric-chau:
Linked issue: #53

mickaelandrieu:
@eric-chau :+1:

eric-chau:
Thank you @crouillon and @mickaelandrieu for your time and the review.

PR #472, opened 2015-06-19
[DependencyInjection] updated the way we handle service that depends on factory

eric-chau:
``Symfony\Component\DependencyInjection\Definition::[get|set]Factory[Method|Class|Service]()`` are now deprecated so this PR updated it to be compliant with this. This also improves:

- Remaining deprecation notices drop from ``43688`` to ``9267``
- Other deprecation notices drop from ``551`` to ``137``


mickaelandrieu:
:+1:  great !

eric-chau:
Thank you @mickaelandrieu.

PR #474, opened 2015-06-19
Fixed Command autoloading for bundles

mickaelandrieu:
``getNamespace()`` is no longer available for Bundles, so we need to reflect classes to get the namespace of Bundle.


mickaelandrieu:
ping @eric-chau  :)

mickaelandrieu:
sorry @eric-chau , need this to continue to work on DebugBundle

PR #486, opened 2015-07-01
[comment] update BBAplication php doc

ndufreche:
type of correction

- Add backslash on complete namespace
- Add comments on uncommented methods
- Return the Interfaces instead of the object him self or the abstract


mickaelandrieu:
The build is broken. Don't worry, I have already provided the fix (https://github.com/backbee/BackBee/pull/487)

PR #488, opened 2015-07-01
Debug toolbar is now available when error page

mickaelandrieu:
When page is in 404 or 500 error, we don't have any information from debug toolbar.

This pull request activate toolbar when we display the Symfony debug response.
debug-toolbar

The UI is not perfect yet but this ask a refactoring of our debug toolbar (which is out of scope of the issue solved by this contribution).

Edit: I've added some styles, we may need a  new UI.
capture2


mickaelandrieu:
ping @eric-chau

PR #489, opened 2015-07-01
[RFR] [NestedNode] added method PageQueryBuilder::andLayoutIs()

cedric-b:
Added missing method ``::andLayoutIs()`` into ``BackBee\NestedNode\Repository\PageQueryBuilder``.

ping @crouillon, is it intentional for you to remove it since this version or you simply forget it?

### DIFF
--- a/NestedNode/Repository/PageQueryBuilder.php
+++ b/NestedNode/Repository/PageQueryBuilder.php
@@ -29,6 +29,7 @@ use Doctrine\ORM\Query\Expr;
 use BackBee\Exception\BBException;
 use BackBee\NestedNode\Page;
 use BackBee\Site\Site;
+use BackBee\Site\Layout;
 
 /**
  * This class is responsible for building DQL query strings for Page.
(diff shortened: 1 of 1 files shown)


ndufreche:
:+1: I have relaunch travis only one instance has faild.

mickaelandrieu:
:+1: but let's wait @crouillon to understand why the method has been removed ;)

crouillon:
Well, yes I intend to get a shorter PageQueryBuilder by removing methods not so useful:
andLayoutIs() is just a shortcut, it as easy and lighter to directly query on _layout when need.
Do you agree?

mickaelandrieu:
@crouillon as a leader you decide, I agree both decisions:
* it's convenient to have some shortcuts
* but the querybuilder classes are very big and hard to maintain and test

Is there a way to define and use our "own" ``PageQueryBuilder`` ?

For instance:

``
/repository
    /QueryBuilder
        /PageQueryBuilder ====> How to declare it to be able to use it
``

```php
<?php
namespace BackBee\Standard\QueryBuilder;

class PageQueryBuilder extends BackBee\NestedNode\QueryBuilder\PageQueryBuilder
{
    // add all shortcuts I need here
}
```

Does this kind of feature be enough to help you @cedric-b ?

cedric-b:
Got bullets (among them "why andSiteIs() then ?") but I don't really care and let you decide about the critical mass of PageQueryBuilder. I just don't wan't to have special pages in conf (routes), cannot find page by root contentset unless default content and prefer something aesthetically nice like: 
            $pages
                ->createQueryBuilder('p')
                ->andLayoutIs($layout)
                ->andSiteIs($this->app->getSite())
                ->andIsOnline()
                ->setMaxResults(1)->getQuery()->getOneOrNullResult();
But it' up to you ping @crouillon :D

crouillon:
:+1: @cedric-b is right :-)

PR #492, opened 2015-07-09
[UrlGenerator] Fix URL generator

crouillon:
Don't take account of deleted pages while checking the uniqueness of an URL to avoid a possible NonUniqueResultException throwed by Doctrine


mickaelandrieu:
:+1:

eric-chau:
nice catch, :+1:

PR #494, opened 2015-07-09
Removed deprecated class String uses from BackBee core

mickaelandrieu:
Bumped utils 2.0 which is compatible with php7 and introduce ``StringUtils`` instead of ``String``.

### DIFF
--- a/Cache/File/Cache.php
+++ b/Cache/File/Cache.php
@@ -27,7 +27,7 @@ use Psr\Log\LoggerInterface;
 
 use BackBee\Cache\AbstractCache;
 use BackBee\Cache\Exception\CacheException;
-use BackBee\Utils\String;
+use BackBee\Utils\StringUtils;
 
 /**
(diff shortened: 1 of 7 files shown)


eric-chau:
Yeah, finally we do it =) thank you @mickaelandrieu :+1:

MaverickWar:
Is this linked to my issue...


Whoops, looks like something went wrong.

1/1
ClassNotFoundException in install.php line 232:
Attempted to load class "StringUtils" from namespace "BackBee\Utils".
Did you forget a "use" statement for "Symfony\Component\Security\Core\Util\StringUtils"?
in install.php line 232

mickaelandrieu:
@MaverickWar what version of BackBee do you use?  Please use the latest version from github or require at least the 1.1 instead or 1.0.

The docs are **still** outdated /c @SylvainLejarre

MaverickWar:
yes I can confirm this is version 1.0

PR #495, opened 2015-07-10
[Cache] Introduce a new CacheInterface

mickaelandrieu:
Added an interface instead of use `AbstractCache` as the contract for each Cache adapter.

ping @crouillon

### DIFF
--- a/Cache/CacheExtendedInterface.php
+++ b/Cache/CacheExtendedInterface.php
@@ -1,0 +1,81 @@
+<?php
+
+/*
+ * Copyright (c) 2011-2015 Lp digital system
+ *
+ * This file is part of BackBee.
+ *
+ * BackBee is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * BackBee is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with BackBee. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * @author Charles Rouillon <charles.rouillon@lp-digital.fr>
+ */
+
+namespace BackBee\Cache;
+
+use Psr\Log\LoggerInterface;
+use BackBee\Cache\Exception\CacheException;
+
+/**
+ * Interface for BackBee Cache Tag & Expire features
+ *
+ * @category    BackBee
+ *
+ * @copyright   Lp digital system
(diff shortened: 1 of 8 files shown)


crouillon:
Good idea, but you'll have to add ExtendedCacheInterface also
The cachelistener should be also changed

Cache/CacheInterface.php
```diff
@@ -0,0 +1,146 @@
...
+    public function setContext($context);
+
+    /**
+     * Logs a message on provided level if a logger is defined.
+     *
+     * @param string $level   The log level
+     * @param string $message The message to log
+     * @param array  $context The logging context
+     */
+    public function log($level, $message, array $context);
```
crouillon: no need this public in interface i think

Cache/CacheInterface.php
```diff
@@ -0,0 +1,146 @@
...
+
+    /**
+     * Returns the expiration timestamp.
+     *
+     * @param int $lifetime
+     *
+     * @return int
+     */
+    public function getExpireTime($lifetime);
```
crouillon: no need this public in interface i think

Cache/CacheInterface.php
```diff
@@ -0,0 +1,146 @@
...
+
+    /**
+     * Control the lifetime against min and max lifetime if provided.
+     *
+     * @param int $lifetime
+     *
+     * @return int
+     */
+    public function getControledLifetime($lifetime);
```
crouillon: no need this public in interface i think

Cache/CacheInterface.php
eric-chau: Agree, the action to log something is up to the developer.

mickaelandrieu:
thanks for review @crouillon

Cache/CacheExtendedInterface.php
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * Interface for BackBee Cache Tag & Expire features
+ *
+ * @category    BackBee
+ *
+ * @copyright   Lp digital system
+ * @author      Mickaël Andrieu <mickael.andrieu@lp-digital.fr>
+ */
+interface CacheExtendedInterface
```
crouillon: should extends CacheInterface

Event/Listener/CacheListener.php
```diff
@@ -133,17 +134,21 @@ public function __construct(ApplicationInterface $application, CacheValidator $v
...
         }
 
-        if (true === $this->application->getContainer()->has('cache.page')) {
-            $cache_page = $this->application->getContainer()->get('cache.page');
-            if (true === ($cache_page instanceof AbstractExtendedCache)) {
-                $this->cache_page = $cache_page;
+        if ($this->application->getContainer()->has('cache.page')) {
+            $cachePage = $this->application->getContainer()->get('cache.page');
+            if ($cachePage instanceof CacheInterface
+                && $cachePage instanceof CacheExtendedInterface
```
crouillon: same as previously

mickaelandrieu:
yes, will be solved in the next pull request :)

PR #496, opened 2015-07-10
[Console] updated every bundle command

eric-chau:
This PR aims to get every bundle command works again. I also moved the ``install`` and ``update`` process from ``AbstractBundle`` to ``BundleLoader`` so every class that extend ``AbstractBundle`` will be lighter.

During the installation or update of a bundle we also dispatch 4 new events: (considering for the example below that the bundle _id_ is ``demo``)

- ``bundle.demo.preinstall`` is dispatched before ``BundleLoader`` install bundle Doctrine entities
- ``bundle.demo.postinstall`` is dispatched after ``BundleLoader`` install bundle Doctrine entities
- ``bundle.demo.preupdate`` is dispatched before ``BundleLoader`` update bundle Doctrine entities
- ``bundle.demo.postupdate`` is dispatched after ``BundleLoader`` update bundle Doctrine entities

There is also two new class events, ``Bundle/Event/BundleInstallEvent`` and ``Bundle/Event/BundleUpdateEvent``.

ping @crouillon, @mickaelandrieu.

### DIFF
--- a/Bundle/BundleLoader.php
+++ b/Bundle/BundleLoader.php
@@ -23,13 +23,17 @@
 
 namespace BackBee\Bundle;
 
-use Symfony\Component\DependencyInjection\Definition;
-use Symfony\Component\DependencyInjection\Reference;
 use BackBee\ApplicationInterface;
 use BackBee\Config\Config;
 use BackBee\DependencyInjection\Util\ServiceLoader;
+use BackBee\Bundle\Event\BundleInstallUpdateEvent;
 use BackBee\Exception\InvalidArgumentException;
 use BackBee\Util\Resolver\BundleConfigDirectory;
+
+use Doctrine\ORM\Tools\SchemaTool;
+
+use Symfony\Component\DependencyInjection\Definition;
+use Symfony\Component\DependencyInjection\Reference;
 
 /**
  * BundleLoader loads and injects bundles into application and its dependency injection container.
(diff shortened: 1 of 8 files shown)


Bundle/BundleLoader.php
```diff
@@ -185,6 +190,143 @@ public function loadBundlesRoutes()
...
+            ->addPaths([
+                $entityDir,
+            ])
+        ;
+
+        $metadata = $bundleEntityManager->getMetadataFactory()->getAllMetadata();
+        $schemaTool = new SchemaTool($bundleEntityManager);
+        $sqls = $schemaTool->getUpdateSchemaSql($metadata, true);
+        if (true === $force) {
+            $schemaTool->updateSchema($metadata, true);
```
mickaelandrieu: why a second parameter ``true`` here and not in others functions ?

eric-chau: Which other method? If you're talking about ``::createSchema()`` about it does not take a second parameter while ``::updateSchema()`` does.

Bundle/BundleLoader.php
```diff
@@ -185,6 +190,143 @@ public function loadBundlesRoutes()
...
+     * Note that if $force is equal to false it will only return updates informations about your bundle but
+     * it won't execute any SQL statement.
+     *
+     * @param  BundleInterface $bundle The bundle to update
+     * @param  boolean         $force  If true it will execute SQL query, default: false
+     * @return array
+     */
+    public function updateBundle(BundleInterface $bundle, $force = false)
+    {
+        $event = new BundleUpdateEvent($bundle, [
```
mickaelandrieu: humm ... I don't know. Do we need empty classes to specify the event ? We already have the information with en dispatched event name.

Console/Command/BundleInstallAllCommand.php
```diff
@@ -57,26 +52,12 @@ protected function configure()
...
-
-            if ($force) {
-                $output->writeln('<info>Running install</info>');
-                $bundle->install();
-            }
-
-            $output->writeln('<info>SQL executed: </info>'.PHP_EOL.implode(";".PHP_EOL, $sqls).'');
-        }
+        return 'install';
     }
```
mickaelandrieu: Install and InstallAll have the same type action ?

Bundle/BundleLoader.php
mickaelandrieu: my bad, misread from my side :)

Bundle/BundleLoader.php
eric-chau: You're right, I think we don't need to be that much precise.

Console/Command/BundleInstallAllCommand.php
eric-chau: Yeah, it's the type of the action, not the action name. I will add some more checks about it.

mickaelandrieu:
:+1:

PR #503, opened 2015-07-15
Rest site controller

ndufreche:
This is an method addition to rest site controller: getSiteCollection.

note: I'm not sure but I think getLayoutsAction method is unused and it can be removed because there is no route to bind this action.

### DIFF
--- a/Config/route.yml
+++ b/Config/route.yml
@@ -387,6 +387,15 @@ bb.rest.layout.delete:
         version: \d+
         _method: DELETE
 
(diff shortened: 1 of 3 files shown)


mickaelandrieu:
This is very important to keep our REST api well tested, we have invested a lot of time of this task.
Can you please add a test on Test/Controller/SiteController ?

ndufreche:
@mickaelandrieu test added.

mickaelandrieu:
nice ! thank you @ndufreche

PR #504, opened 2015-07-15
[RFR] Create a REST v2 for getCollectionAction from PageController

ndufreche:
This PR is Work In Progress and depend of other updates, in particular in backbee/BbCoreJs.

Add new query param "root" and add the possibility ti use getCollectionAction as a real getCollection.

### DIFF
--- a/Rest/Controller/PageController.php
+++ b/Rest/Controller/PageController.php
@@ -114,6 +114,25 @@ class PageController extends AbstractRestController
     /**
      * Get collection of page entity.
      *
+     * Version 1
+     *  - without params return current root
+     *  - parent_uid return first level before the parent page
+     *
+     * Version 2
+     *  - without params return all pages
+     *  - `parent_uid` return all pages available before the nested level
+     *  - `root` return current root
+     *  - `level_offset` permit to choose the depth ex: `parent_uid=oneuid&level_offset=1` equals version 1 parent_uid parameter
+     *  - `has_children` return only pages they have children
+     *  - new available filter params:
+     *    - `title` (is a like method)
+     *    - `layout_uid`
+     *    - `site_uid`
+     *    - `created_before`
+     *    - `created_after`
+     *    - `modified_before`
+     *    - `modified_after`
+     *
      * @return Symfony\Component\HttpFoundation\Response
      *
      * @Rest\Pagination(default_count=25, max_count=100)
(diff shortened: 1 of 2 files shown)


mickaelandrieu:
your actual suite tests should pass or this "means" you have changed actual behavior, we can't allow BC atm

ndufreche:
Yes I have change the actual behavior, because with the actual behavior is impossible to get a clean collection.

The other possibility is create an other route + action but I think soil the Rest Api to stay "BC" is not a better practice.

This is clearly a bug fix for me.

I have update unit test then you can see the new param an the more clean behavior.

mickaelandrieu:
@ndufreche I understand your point but we can't break a REST API: it's far too late, the only viable option is to create another api call with a correct behavior and deprecate the actual for api v2.

Also this does not make sense to update the test in order to valid a new behavior : this way you can valid your tests all the time, test suite is a contract on features, you can change code but never change tests (except if your tests don't test anything).

For me, it's :-1:

ndufreche:
I have update rest version to 2.

Rest/Controller/PageController.php
```diff
@@ -151,6 +151,8 @@ public function getCollectionAction(Request $request, $start, $count, Page $pare
             throw new BadRequestHttpException(
                 'To get page collection by content, you must provide `content_uid` and `content_type` as query parameters.'
             );
+        } elseif ($request->attributes->get('version') == 2) {
```
ndufreche: @eric-chau there is a better solution to implement an version of the REST api ?

ndufreche:
@mickaelandrieu I have done lot of upgrade and add micro and temporally version managing REST version.
Version 1 still totally compatible with the day 1 getCollectionAction response (test have been restored).
Version 2 take lot of improvements and is now a better getCollectionAction implementation (new test have been added).

mickaelandrieu:
There is no perfect solution, I agree to merge this pull request from my side :+1: 

Would you mind to rebase your commits ? /c @eric-chau are you ok ?

ndufreche:
Yes I'm agree, is not the best implementation, in particular the version management. But is the best I can do currently.

mickaelandrieu:
@ndufreche can you squash your commits ? => time to merge : - )

ndufreche:
Ok I do it soon.

ndufreche:
Before squash and merge I will add a last parameter to the V2 
 - `has_children` to return only Sections.

mickaelandrieu:
thank you @ndufreche

ndufreche:
backbee/backbee-standard#326 solve in part with the attribute has_children

Issue #509, opened 2015-07-20
[Bundle Controller] Call of non-existing getter


cedric-b:
The __call function of AbstractAdminBundleController calls the getBundle() method that doesn't exists.
+ The error is not correctly caught, and instead of responding a 401 as intended in the code, it creates an inifinite loop in the __call until maximum loop limit is reached and then returns  a 500....

mickaelandrieu:
ping @ndufreche

mickaelandrieu:
@cedric-b what do you think of https://github.com/backbee/BackBee/pull/510 ?

eric-chau:
@cedric-b, what controller do you extend? Cause we can't find any trace of ``::__call`` in BackBee controller which is the source of your issue.

ndufreche:
This issue show us two problem. We have to improve __call method and Add getBundle method. In addition we can remove the call of getBundle in BackBee core and use directly the bundle property.
- [ ] create getBundle method
- [ ] improve isMethodExists in __call function
- [ ] use bundle property in AbstractAdminBundleController instead of getBundle method

PR #510, opened 2015-07-20
fixed issue 509: added getBundle() to AbstractBundleController

mickaelandrieu:
REFS #509

### DIFF
--- a/Bundle/AbstractBundleController.php
+++ b/Bundle/AbstractBundleController.php
@@ -65,6 +65,16 @@ abstract class AbstractBundleController extends Controller
     }
 
     /**
+     * Get the current bundle
+     *
+     * @param BundleInterface|null $bundle if bundle was not set, return null
+     */
+    public function getBundle()
+    {
+        return $this->bundle;
+    }
+
+    /**
      * Set the current bundle
      *
      * @param BundleInterface $bundle


Bundle/AbstractBundleController.php
```diff
@@ -65,6 +66,24 @@ public function __construct(BBApplication $app)
...
     /**
+     * Get the current bundle
+     *
+     * @param BundleInterface|BundleConfigurationException $bundle if bundle was not set,
+     * return an Exception
+     */
+    public function getBundle()
+    {
+        if (! $this->bundle instanceof BundleInterface) {
+            throw new BundleConfigurationException(
```
ndufreche: Don't throw error if bundle isn't setted, because isn't required.

mickaelandrieu: so ``getBundle`` returns an BundleInterface or null, is that the correct behavior ?

ndufreche: Yes. @eric-chau can validate this behavior ?

eric-chau: I agree with what @ndufreche said especially cause we set the bundle by setter and not the class constructor. So it sounds logical to return ``null`` if not setted.

eric-chau: But I don't really get why we need to retrieve a bundle by controller. Is it really useful?

ndufreche: This is not a system to retrieve a bundle, is just an accessors to help the developer and this only if is automatically setted (currently it work only in the case of an administration bundle controller).

mickaelandrieu:
Such big pull request \o/

eric-chau:
This issue must not exist and it's two distincts issues. First AbstractBundleController has its bundle property as protected so you can access to it from every class that extended it. And the fact that ``::getBundle()`` result on an infinite loop must not be solved by adding the method.

ndufreche:
Yes this resolve just an part of the issue.

mickaelandrieu:
@eric-chau  infinite loop come from __call, if you have getBundle action, this will return null

ndufreche:
Yes but if you called another method non existent, you can reproduce the infinity loop. We have to upgrade __call method to improve it.

eric-chau:
Yeah but this is not a solution to add a getter. Else if you do ``getFooBar`` it will result on an infinite loop too, so to solve this you will create the getter ``getFooBar``? What I mean is that we need to work on ``::__call`` method and not to add the getter. If this getter does not exist before it's because we don't need it.

eric-chau:
Also, I can't find the ``::_call`` method from ``BackBee\Bundle\AbstractBundleController`` or ``BackBee\Controller\Controller`` method, how can @cedric-b get this error? I think he does not extends our controller. This is definitely not BackBee duty to fix this.

mickaelandrieu:
@eric-chau  https://github.com/mickaelandrieu/BackBee/blob/fix-509/Bundle/AbstractBundleController.php#L92

eric-chau:
@mickaelandrieu thank you I didn't notice it.

eric-chau:
So work must be done on this ``::__call`` method and not to add the getter method.

ndufreche:
We need the getter method and the improvement of the __call method.

This PR is correct and ended

mickaelandrieu:
@ndufreche what do I need to do in __call() ?

ndufreche:
Avoid infinite loop and for that rethink the `checkMethodExist` method to fix this problem like check how many action is in the name of the method tested.

eric-chau:
I looked at the ``::__call`` method and this is clearly incomplete. And this method is not useful at all cause we need the bundle in subclass of AbstractBundleController. So why don't you just use ``$this->bundle`` instead of using ``$this->getBundle()``? Plus this will do an useless function call.

ndufreche:
@eric-chau because for implementation, I think is more comprehensive to use an accessors than directly the parameter. Perhaps, in the backbee code we can remove the getter and use the parameter, this is not problematic, but I will keep getBundle method in all cases.

eric-chau:
Ok but to me it's not useful. Nobody will get the bundle from outside of the controller so it will be an unused public method. And inside a subclass it's not optimized to use a getter instead of accessing directly to object property.

ndufreche:
:+1:

PR #511, opened 2015-07-28
[bundle] remove useless function call and overcharging renderer function 

quentin-boulard:
I have replaced '$this->getBundle()' by '$this->bundle' in AbstractAdminBundleController.php

I have add render method for avoid admin templates overcharging behind bundles


eric-chau:
:+1: thank you for your contribution.

Issue #384, opened 2015-05-18
Remove Site MetaData folder


eric-chau:
Seems that ``MetaData`` folder and its classes inside ``Site`` component is out of date.

PR #514, opened 2015-07-30
[Security] fixed session auto logout after n minutes

eric-chau:
This PR brings fix for https://github.com/backbee/backbee-standard/issues/257.

The issue is that the token used by front_area security is setted only once at the first authentication of the user. Rest of the time we keep updating the expire time of the rest_area but not the front_area token.

Ping @crouillon, @mickaelandrieu. I tried to fix it with a better way but our Security component is kinda out of date compare to SF 2.7 recommandations.

### DIFF
--- a/Security/Authentication/Provider/PublicKeyAuthenticationProvider.php
+++ b/Security/Authentication/Provider/PublicKeyAuthenticationProvider.php
@@ -23,13 +23,15 @@
 
 namespace BackBee\Security\Authentication\Provider;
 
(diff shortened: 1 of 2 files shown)


mickaelandrieu:
Let's be pragmatics: if this PR solves the issue,  it's OK to me. 

When we have more time,  we will finish the update of BackBee to Symfony 2.7 components. 

:+1:

eric-chau:
This PR is too specific and not adapted to the needs, replaced by https://github.com/backbee/backbee-standard/pull/384.

PR #518, opened 2015-08-07
get the request from application

mickaelandrieu:
in ExceptionListener.

ping @eric-chau


ndufreche:
:+1:

eric-chau:
Hm why do you get the request object? It seems that you don't use it in the method block.

mickaelandrieu:
@eric-chau I do in the `in_array` to check what kind of Response I need to send.

eric-chau:
@mickaelandrieu my bad I didn't see it. :+1: thank you =)

PR #521, opened 2015-08-11
Corrected link to standard edition

slde-gorillaman:


eric-chau:
thank you @slde-gorillaman :+1:

PR #522, opened 2015-08-12
[ClassContent] Store short content classnames in Db

crouillon:
backbee/backbee-standard#330

### DIFF
--- a/ClassContent/AbstractContent.php
+++ b/ClassContent/AbstractContent.php
@@ -164,23 +164,15 @@ abstract class AbstractContent implements ObjectIdentifiableInterface, Renderabl
      *
      * @return string classname associated to provided
      *
-     * @throws
+     * @throws InvalidContentTypeException Raises if $type is unknown
      */
     public static function getClassnameByContentType($type)
     {
-        $className = self::CLASSCONTENT_BASE_NAMESPACE.str_replace('/', NAMESPACE_SEPARATOR, $type);
-
         try {
-            $exists = class_exists($className);
-        } catch (\Exception $e) {
-            $exists = false;
-        }
-
-        if (!$exists) {
+            return self::getFullClassname(str_replace('/', NAMESPACE_SEPARATOR, $type));
+        } catch (\InvalidArgumentException $e) {
             throw new InvalidContentTypeException($type);
         }
-
-        return $className;
     }
 
     /**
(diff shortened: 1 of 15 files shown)


ClassContent/AbstractContent.php
```diff
@@ -1098,4 +1096,58 @@ private function formatJsonData(array $data, $format)
...
+     * 
+     * @throws InvalidArgumentException     Raises if $classname is not subclass of AbstractClassContent.
+     */
+    public static function getShortClassname($classname)
+    {
+        if (is_object($classname)) {
+            $classname = ClassUtils::getRealClass($classname);
+        }
+
+        if (!is_subclass_of($classname, 'BackBee\ClassContent\AbstractClassContent')) {
```
ndufreche: I'm not sure of put this function in AbstractContent is pertinent if is used only in the `BackBee\ClassContent\AbstractClassContent` scope. In the other scripts you called this function by AbstractClassContent scope => `AbstractClassContent::getShortClassname`

ClassContent/AbstractContent.php
```diff
@@ -1098,4 +1096,58 @@ private function formatJsonData(array $data, $format)
...
+    /**
+     * Returns the full classname of the given $classname.
+     * 
+     * @param  mixed    $classname          A classname or an AbstractClassContent instance, if null the object itself.
+     * 
+     * @return string                       The full classname.
+     * 
+     * @throws InvalidArgumentException     Raises if $classname cannot be resolved as a short classname of AbstractClassContent.
+     */
+    public static function getFullClassname($classname)
```
ndufreche: Same as getShortClassname function

ndufreche:
:+1:

ClassContent/AbstractContent.php
crouillon: thanks @ndufreche, i check it but it must be in AbstractContent because the abstract itself and Revision class have to access it.

I used AbstractClassContent in order to not add a new Use statement, but to not be confusing i can change it, what do you think?

ndufreche: :+1:

mickaelandrieu:
Hi,  thank you @crouillon. 

Some comments:
* can you apply some code style fixes?  You can use phpcs
* can you remove @covers annotations?  this is not required by PHPUnit and may break the coverage command
* does the migration command need to be part of core?  I mean only "legacy projects"  need the update,  how about a migration guide and a dedicated repository with this command? If we add this command we have to deprecate it in 1.1 but maintain it until 2.0.

mickaelandrieu:
thanks!

PR #526, opened 2015-08-20
Add form component as suggested dependency

mickaelandrieu:
If the component is not installed, we are unable to use ``Controller::createFromBuilder()``.

This contribution introduce an exception with an explicit message to solve the issue, and suggest the installation of Symfony Form component on Composer installation.

ping @eric-chau

### DIFF
--- a/Controller/Controller.php
+++ b/Controller/Controller.php
@@ -133,6 +133,10 @@ class Controller implements ContainerAwareInterface
      */
     public function createFormBuilder($data)
     {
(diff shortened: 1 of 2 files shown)


eric-chau:
:+1: but this `BackBee\Controller\Controller` is really painful, we have to remove it asap but it can't be achieve before 2.0 :/

mickaelandrieu:
@eric-chau agree to add some deprecation annotations, but on which methods ? We need to think about this class utility :)

mickaelandrieu:
thanks for review @eric-chau

PR #528, opened 2015-08-21
Classcontent parameters update

crouillon:
- Adding classcontent.postload event to be able to priorize listeners
- Introducing MalformedParameterException to prevent parameters without `value` entry


ClassContent/AbstractClassContent.php
```diff
@@ -997,7 +1009,7 @@ public function getParam($key)
 
         $params = $this->defaultParams[$key];
 
-        if ($value !== null) {
+        if (is_array($value) && isset($value['value'])) {
```
eric-chau: nice catch :+1:

ClassContent/AbstractClassContent.php
```diff
@@ -1031,7 +1043,7 @@ public function getAllParams()
 
         foreach ($this->defaultParams as $key => $value) {
             $params[$key] = $value;
-            if (isset($instanceParams[$key])) {
+            if (isset($instanceParams[$key]) && isset($instanceParams[$key]['value'])) {
```
eric-chau: this condition can be simplified to: ``if (isset($instanceParams[$key]['value'])) {``

Controller/Controller.php
```diff
@@ -133,6 +133,10 @@ public function getEntityManager()
      */
     public function createFormBuilder($data)
     {
+        if (!class_exists('Symfony\\Component\\Form\\Forms')) {
```
eric-chau: Hm I think a rebase is missing? Cause this is not supposed be part of your PR.

Rest/Controller/KeywordController.php
```diff
@@ -21,7 +21,7 @@
  * @author Charles Rouillon <charles.rouillon@lp-digital.fr>
  */
 
-namespace backbee\Rest\Controller;
+namespace BackBee\Rest\Controller;
```
eric-chau: Same for this

Controller/Controller.php
mickaelandrieu: true :)

composer.json
```diff
@@ -84,6 +84,9 @@
         "backbee/bb-core-js": "1.0.0",
         "symfony/phpunit-bridge": "~2.7.0"
     },
+    "suggest": {
```
eric-chau: Same remark for this, must be not part of current PR.

eric-chau:
Good job @crouillon, thank you :+1:

Controller/Controller.php
crouillon: yes but it's rebased, i don't know why this commit appears

mickaelandrieu:
:+1:  thank you @crouillon

PR #529, opened 2015-08-21
[Entity] optimize length field for md5 uids

quentin-boulard:
I rewrite annotation to optimize length field for all uids which are in md5. 
I hope that will optimize memory allocation  for database requests. 

impacted class : 
- ClassContent/AbstractContent.php
- Logging/AdminLog.php
- NestedNode/AbstractNestedNode.php
- NestedNode/KeyWord.php
- NestedNode/MediaFolder.php
- NestedNode/Tests/Mock/MockNestedNode.php 
- Site/Layout.php
- Site/Site.php
- Util/BBAbstract/AbstractUidEntity.php
- Workflow/State.php

I add "length=32," on those fields.

### DIFF
--- a/ClassContent/AbstractContent.php
+++ b/ClassContent/AbstractContent.php
@@ -62,7 +62,7 @@ abstract class AbstractContent implements ObjectIdentifiableInterface, Renderabl
      *
      * @var string
      * @ORM\Id
-     * @ORM\Column(type="string", name="uid")
+     * @ORM\Column(type="string", lenght=32, name="uid")
      */
     protected $_uid;
 
--- a/ClassContent/AbstractContent.php
+++ b/ClassContent/AbstractContent.php
@@ -62,7 +62,7 @@ abstract class AbstractContent implements ObjectIdentifiableInterface, Renderabl
      *
      * @var string
      * @ORM\Id
-     * @ORM\Column(type="string", lenght=32, name="uid")
+     * @ORM\Column(type="string", length=32, name="uid")
      */
     protected $_uid;
 
--- a/Logging/AdminLog.php
+++ b/Logging/AdminLog.php
@@ -42,7 +42,7 @@ class AdminLog
     /**
      * @var string
      * @ORM\Id
-     * @ORM\Column(type="string", name="uid")
+     * @ORM\Column(type="string", length=32, name="uid")
      */
     protected $_uid;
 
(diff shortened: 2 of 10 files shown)


mickaelandrieu:
Hi @quentin-boulard, we probably need a migration command, no ?

quentin-boulard:
No I don't change data just length of the fields in databe so a simple :  
"./backbee bbapp:update" is need

eric-chau:
Good idea @quentin-boulard, thank you :+1:

mickaelandrieu:
@quentin-boulard no risks of database corruption if some data have a length greater than 32 ?

ndufreche:
In reality this have no impact on database size, because from 0 to 255 character mysql use only 1 byte for store the varchar size and after 255 it use 2 bytes [doc link](https://dev.mysql.com/doc/refman/5.0/en/char.html).

And @mickaelandrieu there is no  database corruption in this case of modification.

quentin-boulard:
I checked and I only modified field which are something like that
```
public function __construct($uid = null, $options = null)
    {
        $this->_uid = (is_null($uid)) ? md5(uniqid('', true)) : $uid;
```
so only the uids which are 32 length are set to 32 because it's md5 and md5 always return 32 character

eric-chau:
agree with @quentin-boulard.

mickaelandrieu:
Ok, then :) I prefer ask than regret :p Nice contribution !

ndufreche:
:+1: thanks.

quentin-boulard:
@ndufreche I agree but in my experience we had a big size database with backbee and set the length to 32 on varchar gave us a gain of 5GB on a table only .

I think it's best so set the length if we know exactly the length of the field.

ndufreche:
@quentin-boulard  this not normal even if the uid is a char.
It's possible than put a fix size doctrine transform the column definition from varchar to char and in this case you remove 1 bit by row.
Even in this case it seems impossible to win 5gb on just one table. This size loose is not provoked by a MySQL restart ?

see the table resume into previous doc link on char size storage in MySQL.

quentin-boulard:
@ndufreche maybe, I had just made the set of length and after the alter table the content table was decreased from 17 to 12 Gb, I'm not a pro in doctrine so I suppose that if we set the length we gain space with big volume data

ndufreche:
@quentin-boulard finally what you describe is possible with relations. I'm not sure but the indexation table maybe don't accept varchar as value and store char with length of 255 character.

crouillon:
Thanks @quentin-boulard

PR #530, opened 2015-08-25
removed major and critical issues

mickaelandrieu:


Console/Command/UpgradeToSectionHasChildrenCommand.php
```diff
@@ -162,8 +162,8 @@ private function updateSections()
         $sections = $this->repo->findAll();
         foreach ($sections as $section) {
             PageListener::setSectionHasChildren($this->em, $section);
-            $this->em->flush();
         }
+        $this->em->flush();
```
eric-chau: ping @crouillon  on it, do you see any issue if we did not flush on every loop but only at the end?

ndufreche: This is depending the project

ndufreche:
:+1:

Console/Command/UpgradeToSectionHasChildrenCommand.php
mickaelandrieu: from my tests, nothing wrong happens with this update.

PR #531, opened 2015-08-26
[ClassContent] Introduce AbstractContent::throwExceptionOnUnknownClassname()

crouillon:


ClassContent/AbstractClassContent.php
```diff
@@ -597,7 +598,11 @@ protected function defineData($var, $type = 'scalar', $options = null, $updateAc
                     unset($value);
                 }
             } else {
-                $values[] = ('scalar' == $type) ? '' : new $type(null, $options);
+                try {
+                    $values[] = ('scalar' == $type) ? '' : new $type(null, $options);
```
eric-chau: use `===` instead of `==` please =)

ClassContent/AbstractClassContent.php
```diff
@@ -597,7 +598,11 @@ protected function defineData($var, $type = 'scalar', $options = null, $updateAc
...
+                } catch (ClassNotFoundException $ex) {
+                    self::onUnknownClassname(sprintf('Unknwon classname `%s`.', $type), $ex);
```
eric-chau: Unknown*

ClassContent/AbstractContent.php
```diff
@@ -50,6 +50,14 @@
     const CLASSCONTENT_BASE_NAMESPACE = 'BackBee\ClassContent\\';
 
     /**
+     * If FALSE (default) raises an exception on unknown class content classname.
+     * If set to TRUE, remind silent
+     * 
+     * @var boolean
+     */
+    private static $ignoreUnknownClassname = false;
```
eric-chau: can you move this declaration after declaration of constants? So we can easily read and find out informations about constant, static and instance attributes. Recap:
```
class FooBar
{
    // constant declaration
    // then static attribute declaration (public => protected => private)
    // then instance attribute declaration (public => protected => private)

    const RANDOM_CONSTANT = 0;

    public static $randomPublicStatic;
    protected static $randomProtectedStatic;
    private static $randomPrivateStatic;

    public $randomPublicAttribute;
    protected $randomProtectedAttribute;
    private $randomPrivateAttribute;
}
```

ClassContent/AbstractContent.php
```diff
@@ -1151,10 +1159,38 @@ public static function getFullClassname($classname)
...
+        return null;
+    }
+
+    /**
+     * If set to FALSE (default) raises an exception on unknown class content classname.
+     * If set to TRUE, remind silent.
+     * 
+     * @param boolean $ignoreUnknownClassname Optional
+     */
+    public static function ignoreUnknownClassname($ignoreUnknownClassname = false)
```
eric-chau: Just a notice, to be really consistent about the method name and the default value of `$ignoreUnknownClassname` the default value of the parameter must be `true` and not `false`.

ClassContent/Listener/ClassContentListener.php
```diff
@@ -274,4 +274,22 @@ public static function onPostCall(Event $event)
...
+    /**
+     * Occurs on event ``bbapplication.init`` to set AbstractContent::$ignoreUnknownClassname according to bbapp parameters.
+     *
+     * @param Event $event
+     */
+    public static function onApplicationInit(Event $event)
+    {
+        $application = $event->getTarget();
+        $container = $application->getContainer();
+        $parameter = $container->getParameter('bbapp.classcontent.ignore_unknown_classname');
```
eric-chau: You must test that the container contains `bbapp.classcontent.ignore_unknown_classname` parameter first.

ClassContent/Listener/ClassContentListener.php
```diff
@@ -274,4 +274,22 @@ public static function onPostCall(Event $event)
 
         $response->setContent(json_encode($content));
     }
+
+    /**
+     * Occurs on event ``bbapplication.init`` to set AbstractContent::$ignoreUnknownClassname according to bbapp parameters.
+     *
+     * @param Event $event
+     */
+    public static function onApplicationInit(Event $event)
```
eric-chau: Not sure that we should include this listener inside BackBee core, what do you think about it?

eric-chau:
Nice contribution @crouillon but i'm not really in about that this listener inside BackBee core, it still very acceptable.

ClassContent/AbstractContent.php
crouillon: Oups didn't notice other constants...you're right

ClassContent/Listener/ClassContentListener.php
crouillon: Well, it can be seen as an in-core feature of BackBee as debug mode or even cache directory choice: a default behavior can easily be overrided by editing yml files without havng to define new event or listener.

eric-chau: Yeah after thought I agree about that this is perfectly legit to provide this feature inside BackBee core. Forget my notice!

ClassContent/Listener/ClassContentListener.php
eric-chau: So this comment can be ignored according to our talks about the listener.

ClassContent/AbstractContent.php
crouillon: Quite tricky: the default value have to be false for compatibilty, but notIgnoreUnknownClassname or raiseExceptionOnUnknownClassname or whatever i thank about to is too long.
If you have any good idea, i take it

ClassContent/AbstractContent.php
```diff
@@ -1151,10 +1159,38 @@ public static function getFullClassname($classname)
...
+        if (!self::$ignoreUnknownClassname) {
+            throw new \InvalidArgumentException($message, 0, $exception);
+        }
+
+        return null;
+    }
+
+    /**
+     * If set to FALSE (default) raises an exception on unknown class content classname.
+     * If set to TRUE, remind silent.
```
eric-chau: If setted

ClassContent/AbstractContent.php
eric-chau: What about renaming the method to `public function throwExceptionOnUnknownClassname($throwException = true)`? So the default behavior would remain the same.

ClassContent/AbstractContent.php
mickaelandrieu: I'm not sure, never seen "setted"

eric-chau: set is not an irregular verb =) type it in google or any translator tools ;)

eric-chau:
:+1: thank you @crouillon

PR #532, opened 2015-08-26
[Config] fix case for getaction.postcall in events.yml

quentin-boulard:
I change getAction to getaction because all events are in lower case in events.yml and in ClassContent/Test/Listener


eric-chau:
:+1: nice catch thank you @quentin-boulard

ndufreche:
:+1:

PR #533, opened 2015-08-26
[Rest] Update GroupController::getCollectionAction()

crouillon:
GroupController::getCollectionAction() must return groups of the current site AND common groups (ie with $_site null)

### DIFF
--- a/Rest/Controller/GroupController.php
+++ b/Rest/Controller/GroupController.php
@@ -57,7 +57,14 @@ class GroupController extends AbstractRestController
             $site_uid = $this->getApplication()->getSite()->getUid();
         }
 
-        $groups = $this->getEntityManager()->getRepository('BackBee\Security\Group')->findBy(['_site' => $site_uid]);
+        $groups = $this->getEntityManager()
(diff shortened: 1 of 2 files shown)


ndufreche:
:+1:

Rest/Controller/GroupController.php
```diff
@@ -57,7 +57,14 @@ public function getCollectionAction(Request $request)
...
 
-        $groups = $this->getEntityManager()->getRepository('BackBee\Security\Group')->findBy(['_site' => $site_uid]);
+        $groups = $this->getEntityManager()
+                ->getRepository('BackBee\Security\Group')
+                ->createQueryBuilder('g')
+                ->where('g._site = :siteUid')
+                ->orWhere('g._site IS NULL')
+                ->setParameter(':siteUid', $site_uid)
+                ->getQuery()
+                ->getResult();
```
eric-chau: Maybe we should consider to create a method inside GroupRepository instead of building the query in controller?

eric-chau:
:+1:

PR #544, opened 2015-09-14
[ClassContent] Fix ClassContentRepository::getLastByMainnode()

crouillon:
No content were found by this method because of unknown classname


eric-chau:
nice catch @crouillon, :+1:

PR #548, opened 2015-09-16
[REST] Add current page on get action as html for classcontent

fkroockmann:
Currently we need to have a current page in classcontent when we ask a new render of content


crouillon:
:+1:

eric-chau:
Closed cause this already works fine.

PR #549, opened 2015-09-18
[ClassContent] updated ClassContentManager::prepareElements()

eric-chau:
Updated `ClassContentManager::prepareElements()`  to handle array of values of an element. Example, if you want to add two keywords to an article:

```js
// PUT /rest/1/classcontent/Article/Aticle/d775e20c62b018d84c1a12229ef9477d
{
    "elements": {
        "keywords": [
            {
                "uid": "04b89c815c65bcde9be3b5fb28f72f64",
                "type": "Element/Keyword"
            },
            {
                "uid": "dee512eb518042412ea374f2e8c272e8",
                "type": "Element/Keyword"
            }
        ]
    },
    "type": "Article/Article",
    "uid": "d775e20c62b018d84c1a12229ef9477d"
}
```

ping @crouillon, @fkroockmann.

### DIFF
--- a/Rest/Controller/ClassContentController.php
+++ b/Rest/Controller/ClassContentController.php
@@ -226,8 +226,6 @@ class ClassContentController extends AbstractRestController
      */
     public function putAction($type, $uid, Request $request)
     {
-        $this->granted('VIEW', $this->getClassContentManager()->findOneByTypeAndUid($type, $uid, true));
-        $this->granted('EDIT', $this->getClassContentManager()->findOneByTypeAndUid($type, $uid, true));
         $this->updateClassContent($type, $uid, $request->request->all());
         $this->getEntityManager()->flush();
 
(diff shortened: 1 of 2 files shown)


Rest/Controller/ClassContentController.php
```diff
@@ -227,7 +227,6 @@ public function postAction($type, Request $request)
     public function putAction($type, $uid, Request $request)
     {
         $this->granted('VIEW', $this->getClassContentManager()->findOneByTypeAndUid($type, $uid, true));
-        $this->granted('EDIT', $this->getClassContentManager()->findOneByTypeAndUid($type, $uid, true));
```
crouillon: Hum, EDIT right seems required to put a content, isn't it?

eric-chau: @crouillon yes but this is already checked in the method `::updateClassContent()` so I removed it to prevent duplicate check.

mickaelandrieu: agree as ``prepareElements`` is private and can't be accessed from outside: only public methods should be checked by ACL from my point of view.

mickaelandrieu:
:+1:

crouillon:
:+1:

PR #552, opened 2015-09-22
[Site] Add Serializer\Type for Site label and layouts

quentin-boulard:
I had some error in deserialize some contents and correct it with adding serializer type.


mickaelandrieu:
:+1:  do you have time to complete a test to check the rendering from REST API comply to your needs ? 
Maybe here ? => https://github.com/backbee/BackBee/blob/master/Rest/Tests/Controller/SiteControllerTest.php#L178

quentin-boulard:
Sorry but no I don't have time for this. I just add Serializer\type and name because you set the @Serializer\Expose without.

mickaelandrieu:
It's not up to me to merge pull requests without tests :/

ping @eric-chau, @crouillon you decide.

eric-chau:
:+1: ok @quentin-boulard. Before I merge it, can you tell what you get in layouts key of serialized site? Does the serializer also call serializer on layouts? In others words does layout data get well serialized too?

PR #553, opened 2015-09-22
introduced getResourcesDirectory() in AbstractBundle

mickaelandrieu:
As discussed with some of the core team, we need this accessor to get translation files from Bundles.

Sorry for tests, there is no one I will try to cover the Bundle part in another pull request.


eric-chau:
:+1:

PR #554, opened 2015-09-22
[ClassContent] updated AbstractContent::computeElementsToJson() 

eric-chau:
Updated ``AbstractContent::computeElementsToJson()`` to always return elements key whatever its values.

ping @crouillon, @fkroockmann.


crouillon:
:+1:

PR #555, opened 2015-09-22
[Event] Fix wrong status code on KernelException

crouillon:


eric-chau:
:+1: thank you @crouillon finally someone did it haha

mickaelandrieu:
It's mainly a hack and shouldn't be solved in listener but in Front controller (or in the another ExceptionListener where all theses exceptions -  with invalid HTTP Status code - are thrown). 

I didn't do it because now we have hidden the issue this won't be fixed,  ever :/

eric-chau:
@mickaelandrieu to me this is not an hack cause both are part of BackBee (I mean `FrontController` and `ExceptionListener`). I'm kinda fine with the fact that BackBee EventListener handle its FrontController exception. Is it really worth to declare another listener for that? I'm not sure. We have already plenty of listeners.

mickaelandrieu:
@eric-chau it's not up to a listener to fix a bad behavior of our own exceptions :/
But it's merged and so on the debate is closed ;)

eric-chau:
@mickaelandrieu it's not a bad behavior, this is working like it is supposed to do. What make you think that this is not the case? :o

mickaelandrieu:
@eric-chau because theses exceptions return bad status code like 60500, 60404 instead of 500, 404 :)

PR #560, opened 2015-09-25
[Site] add missing annotation and remove excess annotation

quentin-boulard:
I add missing annotation and remove excess annotation

### DIFF
--- a/Rest/Tests/Controller/SiteControllerTest.php
+++ b/Rest/Tests/Controller/SiteControllerTest.php
@@ -202,10 +202,14 @@ class SiteControllerTest extends TestCase
         $controller = $this->getController();
         $serialized = $controller->formatItem($this->site);
         $content = json_decode($serialized, true);
-        $this->assertInternalType('array', $content);
+
+        $this->assertInternalType('string', $content['label']);
+        $this->assertEquals($this->site->getLayouts()->count(), count($content['layouts']));
+
+        foreach ($this->site->getLayouts() as $key => $layout) {
+            $this->assertSame(json_decode($controller->formatItem($layout), true), $content['layouts'][$key]);
+        }
     }
-
-
 
     protected function tearDown()
     {
--- a/Rest/Tests/Controller/SiteControllerTest.php
+++ b/Rest/Tests/Controller/SiteControllerTest.php
@@ -197,6 +197,16 @@ class SiteControllerTest extends TestCase
         $this->assertCount(1, $content);
     }
 
+    public function test_serializeSite()
+    {   
+        $controller = $this->getController();
+        $serialized = $controller->formatItem($this->site);
+        $content = json_decode($serialized, true);
+        $this->assertInternalType('array', $content);
+    }
+
+
+
     protected function tearDown()
     {
         $this->dropDb($this->getBBApp());
--- a/Site/Site.php
+++ b/Site/Site.php
@@ -75,7 +75,6 @@ class Site extends AbstractObjectIdentifiable
      * @BB\Fixture(type="domainWord")
      *
      * @Serializer\Expose
-     * @Serializer\SerializedName("label")
      * @Serializer\Type("string")
      *
      */
@@ -126,8 +125,7 @@ class Site extends AbstractObjectIdentifiable
      * @ORM\OneToMany(targetEntity="BackBee\Site\Layout", mappedBy="_site", fetch="EXTRA_LAZY")
      *
      * @Serializer\Expose
-     * @Serializer\SerializedName("layouts")
-     * @Serializer\Type("ArrayCollection<BackBee\Site\Layout>")
+     * @Serializer\Type("ArrayCollection<string, BackBee\Site\Layout>")
      */
     protected $_layouts;
 
--- a/Site/Site.php
+++ b/Site/Site.php
@@ -75,6 +75,9 @@ class Site extends AbstractObjectIdentifiable
      * @BB\Fixture(type="domainWord")
      *
      * @Serializer\Expose
+     * @Serializer\SerializedName("label")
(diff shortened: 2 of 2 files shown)


mickaelandrieu:
Hi,
we cant remove an annotation, at this break the API.

Also, again, 0 tests: how can we be sure that theses contributions still works as expected ?

quentin-boulard:
like I said in description and if you look a the code you ll see that I remove an exceed 
@Serializer\Expose which are twice in this definition.

0 tests? currently you don't have test for that in SiteController and you have errors...

you can write a test for serialization and you must see that without my change you have error for label and layouts 

if you put @Serializer\Expose to a variable you must set a @Serializer\Type

So write your tests and you ll see that without my add/remove you ll have errors.

mickaelandrieu:
except if I'm wrong, all the REST API is well tested, see https://github.com/backbee/BackBee/blob/master/Rest/Controller/SiteController.php for tests related to the availables queries on SIte entity.

Thanks to your own test, that prove this pull request is actualy incomplete as it does not comply what you expect to do.

mickaelandrieu:
@quentin-boulard fixup your commits please, I will merge /c @eric-chau

eric-chau:
Just saying but your test is not really useful cause you test that if ``json_decode()`` can return an array, ofc it does =)

quentin-boulard:
@mickaelandrieu unfortunately you are wrong. You don't test serialization / deserialization for your entities.

* here you check for a MockUser https://github.com/backbee/BackBee/blob/master/Rest/Tests/Controller/AbstractRestControllerTest.php but you set @Serializer\Expose on many class but never test... like :


* /NestedNode/KeyWord.php
* /NestedNode/Page.php
* /Security/Group.php
* /Security/User.php
* /Site/Layout.php
* /Site/Site.php
* /Workflow/State.php


if you make serialized/deserialized  test you will see that you have error in several class which don't have @Serializer\Type and it's needed for deserialize function so if someone try to use  @Serializer to deserialize one of this three classes for a dev or something, he will have an error.


* /Security/User.php
    * $_groups 
* /Security/Group.php 
    * $_id
    * $_name
    * $_description
    * $_users
* /Site/Site.php
    * $_label
    * $_layouts

mickaelandrieu:
@quentin-boulard correct.  What do you suggest?

quentin-boulard:
@eric-chau correct.

@mickaelandrieu add a @Serializer\Type to this different class annotation and write correct generics tests for serialization / deserialization and apply to all classes with Serializer annotation.

Rest/Tests/Controller/SiteControllerTest.php
```diff
@@ -197,6 +197,16 @@ public function test_getSiteController()
         $this->assertCount(1, $content);
     }
 
+    public function test_serializeSite()
+    {   
+        $controller = $this->getController();
+        $serialized = $controller->formatItem($this->site);
+        $content = json_decode($serialized, true);
+        $this->assertInternalType('array', $content);
```
mickaelandrieu: ping @eric-chau can you help him on that point ? If the feature works as expected this should be testable, but maybe with BackBee internal functions ? I don't know this part of BackBee.

mickaelandrieu:
@quentin-boulard I have created an issue / to your last comment. I can't say when it will be done /c @clementbonfils .

Sorry for the time we spend on it, but it's important for the future: BackBee won't be BackBuilder, we try to make a good product, and limit bugs (when possible).

Thanks for your understanding :)

eric-chau:
Hello,

that how you should write your test:
```php
public function test_serializeSite()
{
    $controller = $this->getController();
    $serialized = $controller->formatItem($this->site);
    $content = json_decode($serialized, true);

    $this->assertInternalType('string', $content['label']);
    $this->assertEquals($this->site->getLayouts()->count(), count($content['layouts']));
    foreach ($this->site->getLayouts() as $key => $layout) {
        $this->assertSame(json_decode($controller->formatItem($layout), true), $content['layouts'][$key]);
    }
}
```
Also, can you add this annotation on `BackBee\Site\Site` `_layouts` property please? `@Serializer\SerializedName("layouts")`

Fixup your commits and we will merge it ;) thank you for your time =)

quentin-boulard:
Ok, we need deserialization tests too no ?

eric-chau:
@quentin-boulard to me you don't need to test the deserialization we don't use it at all.

quentin-boulard:
@eric-chau you maybe not but it's like that I saw the problem. when I need to deserialize a Site I had : 

 "[JMS\Serializer\Exception\RuntimeException]             
  You must define a type for BackBee\Site\Site::$_label."  
and 
"[JMS\Serializer\Exception\RuntimeException]             
  You must define a type for BackBee\Site\Site::$_layouts."

so even if you don't use it now, if you set Serializer you are likely to use deserialization as well, no ?

eric-chau:
@quentin-boulard, using JMS deserialize is a bad idea cause the entity won't be managed by Doctrine and this one of plenty of potential issue. Instead of deserialize and get your object, use something like ``$this->em->find('BackBee\Site\Site', $content['uid'])``. By doing it this way you will reduce drastically your chance to meet problem.

Finally, if you still really want to use JMS deserialize, feel free to write the right test for it =)

quentin-boulard:
@eric-chau thanks for this, I just base this on : https://github.com/backbee/BackBee/blob/master/Rest/Controller/AbstractRestController.php#L136-L147
but if you say it's better to use doctrine I ll try.

eric-chau:
You're welcome. I didn't even know that this method exists haha.

quentin-boulard:
this is a big project you can't know all methods in the project :+1:

eric-chau:
thank you @quentin-boulard :+1:

PR #562, opened 2015-09-28
[ClassContent] on content commit we force autoload of classcontent classes

eric-chau:
This PR will fix https://github.com/backbee/backbee-standard/issues/440.

To be concise, in some cases Doctrine doesn't know every classcontent classnames while they are used as factors in request to database. So we force the autoloading of it to avoid this issue.


crouillon:
:+1:

PR #564, opened 2015-09-29
Fix bug objectKeyword does not a parameter

supernana:
I have noticed than all parameters must be set in ClassContent (in lower case).
That PR fixed this problem for ClassContent/Element/Keyword

- Add parameters/objectkeyword in ClassContent/Element/Keyword as array
- Change name from 'objectKeyword' to 'objectkeyword' in NestedNode/Repository/KeyWordRepository

### DIFF
--- a/ClassContent/Element/Keyword.yml
+++ b/ClassContent/Element/Keyword.yml
@@ -7,3 +7,5 @@ Keyword:
     value: !!scalar
   fixture:
     value: word
+  parameters:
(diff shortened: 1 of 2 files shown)


mickaelandrieu:
Hi @supernana,
this make sense :+1:

mickaelandrieu:
thanks

supernana:
thanks

PR #570, opened 2015-10-05
[ClassContent] Disabling cache in ClassContentManager::getAllClassContentClassnames() when debug mode is enabled

crouillon:


eric-chau:
Nice addition, thank you @crouillon

PR #571, opened 2015-10-05
[EntityFinder] Fix namespace resolution of a php class file

crouillon:


eric-chau:
:+1: thank you @crouillon

PR #575, opened 2015-10-07
[Config] Fix default return of method Configurator:getConfigDefaultSections()

crouillon:


eric-chau:
nice catch @crouillon thank you :+1:

mickaelandrieu:
:+1:

PR #576, opened 2015-10-09
Fix stat for image upload

fkroockmann:
If the file are upload on remote server, the stat function can not be used.


crouillon:
:+1:

eric-chau:
:+1: nice catch thank you @fkroockmann

PR #577, opened 2015-10-09
[ClassContentManager] ClassContent directory should not be required in context

crouillon:


eric-chau:
:+1: nice catch @crouillon =)

PR #579, opened 2015-10-14
Allow to add multiple uid as criteria for classcontent::getCollection

fkroockmann:
Now we can :

/res/1/classcontent?uids=foo,bar

### DIFF
--- a/ClassContent/Repository/ClassContentQueryBuilder.php
+++ b/ClassContent/Repository/ClassContentQueryBuilder.php
@@ -77,9 +77,12 @@ class ClassContentQueryBuilder extends QueryBuilder
         if ($site instanceof Site) {
             $site = $site->getUid();
         }
-        $this->andWhere(
-            'cc._uid IN (SELECT i.content_uid FROM BackBee\ClassContent\Indexes\IdxSiteContent i WHERE i.site_uid = :site_uid)'
-        )->setParameter('site_uid', $site);
+
+        if (!empty($site)) {
+            $this->andWhere(
+                'cc._uid IN (SELECT i.content_uid FROM BackBee\ClassContent\Indexes\IdxSiteContent i WHERE i.site_uid = :site_uid)'
+            )->setParameter('site_uid', $site);
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


crouillon:
:+1:

ClassContent/Repository/ClassContentQueryBuilder.php
```diff
@@ -77,9 +77,12 @@ public function addSiteFilter($site)
         if ($site instanceof Site) {
             $site = $site->getUid();
         }
-        $this->andWhere(
-            'cc._uid IN (SELECT i.content_uid FROM BackBee\ClassContent\Indexes\IdxSiteContent i WHERE i.site_uid = :site_uid)'
-        )->setParameter('site_uid', $site);
+
+        if (!empty($site)) {
```
eric-chau: To keep things simple, can you use an extra variable? Something like that:

```php
$uid = null;
if ($site instanceof Site) {
    $uid = $site->getUid();
}

if (null !== $uid) {
    // do your stuff
}
```

fkroockmann: This function accept uid or instance of Site\Site

Rest/Controller/ClassContentController.php
```diff
@@ -652,6 +652,10 @@ private function findContentsByCriteria(array $classnames, $start, $count)
         unset($criterias['order_by']);
         unset($criterias['order_direction']);
 
+        $criterias['contentIds'] = $this->computeUids();
```
eric-chau: I don't think that you need to do it in a function. What do you think about something like that? 
```php
$criterias['contentIds'] = explode(',', $this->getRequest()->query->get('uids', ''));
```
So you can remove ``::computeUids()``.

ClassContent/Repository/ClassContentQueryBuilder.php
eric-chau: oh ok. Nevermind then.

eric-chau:
:+1:

Rest/Controller/ClassContentController.php
fkroockmann: Right :)

PR #580, opened 2015-10-15
Fix classcontent getcollection call

fkroockmann:
Explode return an non empty array, add array_filter for remove this case


eric-chau:
:+1: nice catch, I will merge it when it's all green.

fkroockmann:
thanks :)

PR #581, opened 2015-10-16
adding exception to KeywordController

hbaptiste:
adding exception to KeywordController


eric-chau:
:+1:, thank you @hbaptiste

mickaelandrieu:
Omg we really need to discuss on Monday :/

1) exceptions are already returned as json responses from the api: there is no need to create a private function for that in this place. 

2) the error message returned doesn't help developper and afaik can't be really handled by javascript client.

PR #582, opened 2015-10-16
removing ACL for KwController

hbaptiste:
removing ACL for KwController


eric-chau:
:+1: nice catch =)

mickaelandrieu:
To be honest I don't see why this controller shouldn't be handled by rights like any others resources :/

If rights are missing,  the rights configuration should be updated according to each available role.

PR #583, opened 2015-10-19
Catch RuntimeExcepion for mailer not available

fkroockmann:
https://github.com/backbee/backbee-standard/issues/492


mickaelandrieu:
:+1:

PR #584, opened 2015-10-20
Add extra for add special config in elements part of yml

fkroockmann:
For example we can set the accept of upload for file like:

```
elements:
        file:
            type: BackBee\ClassContent\Element\File
            label: fichier
            extra:
                dropzone:
                    acceptedFiles: '.pdf'
```


mickaelandrieu:
:+1:

PR #587, opened 2015-10-21
Fix Resource dir

fkroockmann:
Currently if the addResourceDir is call before getResourceDir, the resource dirs are not initialize


eric-chau:
Nice addition :+1: thank you @fkroockmann

PR #588, opened 2015-10-21
Fix if draft dont have content

fkroockmann:
Fix if draft dont have content


eric-chau:
:+1: @fkroockmann, thank you.

PR #589, opened 2015-10-21
fixing pagination listener

hbaptiste:
fixing pagination listener


eric-chau:
:+1: nice catch @hbaptiste

PR #592, opened 2015-10-23
fixing pagination eventListener

hbaptiste:
fixing pagination eventListener


eric-chau:
Nice catch @hbaptiste! :+1:

PR #593, opened 2015-10-26
[Bundle] Define BundleLoader as dumpable service

crouillon:
Allow to dump then restore the bundleInfos array allowing to find bundle Id from its path


mickaelandrieu:
:+1:

PR #596, opened 2015-10-27
[Keyword] Update Keyword to Content joins

crouillon:
Something forgotten I think


mickaelandrieu:
:+1: thank you @crouillon

PR #597, opened 2015-10-29
fix clone action

mickaelandrieu:
I don't know why but sometimes it's an instance of PageRepository that is returned as subcontent of content.

ping @eric-chau


eric-chau:
:+1:

PR #598, opened 2015-10-30
[ClassContent] updated ::getAction() in REST controller 

eric-chau:
Currently we cannot re-render content on dynamical page cause the page uid is required by ``Rest\Controller\ClassContentController::getAction`` while this should not be.


crouillon:
:+1: thx @eric-chau

PR #599, opened 2015-10-30
[ClassContent] updated AbstractClassContent::defineParam()

eric-chau:
This PR aims to allow subclass to override parent class parameters.


crouillon:
:+1:

PR #600, opened 2015-11-02
[Perf] Added micro optimizations

mickaelandrieu:
This little patch make some functions more cacheable by the PHP compiler because it know that they are  native functions and not from the class they were used.

In some cases, PHP optimize them (is_*, defined(), strlen). As this patch can't break anything and may improve the performances, let's do it (http://www.slideshare.net/nikita_ppv/php-7-what-changed-internally-php-barcelona-2015).


fkroockmann:
I not agree for add "\" before each native function :-1:

eric-chau:
:+1:

mickaelandrieu:
ping @crouillon what do you think ?

mickaelandrieu:
branch conflicted:feel free to re-apply this improvements.

PR #602, opened 2015-11-02
[ClassContent] updated Element\KeywordRepository::updateKeywordLinks()

eric-chau:
updated ``BackBee\ClassContent\Repository\Element\KeywordRepository::updateKeywordLinks()`` to put third argument as optional.


crouillon:
:+1:

PR #603, opened 2015-11-02
[POC] Http Cache naive implementation

mickaelandrieu:
Hi,

this is a first naive implementation of Http cache using cache control system in BackBee.

Home page of BackBee standard without cache => 1.3s
Hoe page of BackBee standard with http cache => 119ms

On i7, 8Go and SSD.

Let's discuss now on how we should provide this kind of feature in CMS.


fkroockmann:
Nice job dude

mickaelandrieu:
I can probably suggest a complete implementation, *but* I want to be sure this will be merged.

Tell me.

eric-chau:
@mickaelandrieu, you have a good approach right here but I think it will require alot of work to make it works globally.

eric-chau:
We will talk deeper about this topic. Thank you @mickaelandrieu for your contribution.

PR #607, opened 2015-11-06
[POC] disable xdebug

mickaelandrieu:
we don't neex xdebug because we dont use code coverage


mickaelandrieu:
ping @crouillon :)

PR #608, opened 2015-11-09
Make the session configuration optional

mickaelandrieu:
Please note that the `bb_session` service is configured by `session` key /c @fkroockmann @eric-chau @crouillon


eric-chau:
:+1:

PR #614, opened 2015-11-10
Fix retrieve of data when an element is null

fkroockmann:
If an element is null, we need to retrieve a fake element


eric-chau:
:+1:

PR #615, opened 2015-11-10
Updated the PdoSessionHandler configuration

mickaelandrieu:
Avoid the lock of the session when the session is persisted by the same connection as the application.


eric-chau:
:+1:

mickaelandrieu:
ty for review @eric-chau

PR #617, opened 2015-11-10
Refactoring metadata computing

crouillon:
- Introduce MetaDataResolver as service
- Move MetaDataListener in package

### DIFF
--- a/MetaData/Listener/MetaDataListener.php
+++ b/MetaData/Listener/MetaDataListener.php
@@ -1,0 +1,120 @@
+<?php
+
+/*
+ * Copyright (c) 2011-2015 Lp digital system
+ *
+ * This file is part of BackBee.
+ *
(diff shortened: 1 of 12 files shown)


MetaData/Listener/MetaDataListener.php
```diff
@@ -21,11 +21,12 @@
  * @author Charles Rouillon <charles.rouillon@lp-digital.fr>
  */
 
-namespace BackBee\Event\Listener;
```
mickaelandrieu: I don't think you can remove this class, because this create a BC, you can only deprecate it if you want to comply semver.

crouillon:
Event\Listener\MetaDataListerner.php restored for BC reason.
Thanks @mickaelandrieu

mickaelandrieu:
:+1:

PR #618, opened 2015-11-13
Fix override of config to merge several config instead of replace

fkroockmann:
Currently several same config is not merged.

Wait merge of https://github.com/backbee/utils/pull/5


eric-chau:
:+1:

PR #619, opened 2015-11-16
Add tests on entityManager validity

crouillon:
Allow BackBee to start without doctrine.yml


eric-chau:
Nice catch @crouillon, this is really better than before. :+1:

PR #622, opened 2015-11-24
Fix keyword

mickaelandrieu:
this is an amzing fix


BBApplication.php
```diff
@@ -511,15 +511,17 @@ public function getBBConfigDir()
      */
     public function getEntityManager($name = 'default')
     {
-        try {
-            if (null === $this->getContainer()->get('doctrine')) {
+        if ($this->getContainer()->getDefinition('em')->isSynthetic()) {
+            try {
                 $this->initEntityManager();
-            }
```
mickaelandrieu: ceci ne me convient 

changes sts

PR #623, opened 2015-11-30
Fix bug when  is null like launch with jenkins

supernana:
I have notice this line could make some troubles with twig cache

### DIFF
--- a/Renderer/Adapter/Twig.php
+++ b/Renderer/Adapter/Twig.php
@@ -77,7 +77,7 @@ class Twig extends AbstractRendererAdapter
         if ($application->isDebugMode() || (isset($config['debug']) && true === $config['debug'])) {
             $this->twig->enableDebug();
             $this->twig->addExtension(new \Twig_Extension_Debug());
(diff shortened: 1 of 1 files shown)


mickaelandrieu:
:+1:

Renderer/Adapter/Twig.php
```diff
@@ -77,7 +77,7 @@ public function __construct(AbstractRenderer $renderer, array $config = [])
         if ($application->isDebugMode() || (isset($config['debug']) && true === $config['debug'])) {
             $this->twig->enableDebug();
             $this->twig->addExtension(new \Twig_Extension_Debug());
-        } elseif ($application->isClientSAPI()) {
+        } elseif ( null !== $application && false === $application->isClientSAPI()) {
```
eric-chau: So you inverted the ``$application->isClientSAPI()``? Is it really what you wanted to do? Also, you test if ``$application`` is not null, wouldn't it be better to do it earlier? As the first ``if`` condition can benefit of it too.

eric-chau:
Outdated PR.

PR #628, opened 2015-12-10
Added security.interactive_login event

mickaelandrieu:
Dispatch ```security.interactive_login`` event when an user is authentified.


crouillon:
:+1: before the update of this package according to the security component of Symfony

PR #633, opened 2015-12-17
preserve FrontControllerException in TwigAdapter

gobjila:


Renderer/Adapter/Twig.php
```diff
@@ -177,7 +177,15 @@ public function renderTemplate($filename, array $templateDir, array $params = []
         try {
             $params['this'] = $this;
             $params = array_merge($params, $vars);
-            $render = $this->twig->render($filename, $params);
+
+            try {
+                $render = $this->twig->render($filename, $params);
+            } catch (\Twig_Error_Runtime $e) {
+                if($e->getPrevious() instanceof FrontControllerException) {
+                    throw $e->getPrevious();
```
mickaelandrieu: I'mp not sure this is a good idea, mainly because the exception from ``FrontController` is irelevant to help you figure out why Twig can't render properly your template.

eric-chau: Well this PR lacks of description.

mickaelandrieu: You may set the correct status code and keep the relevant message exception I guess.  As you like :)

crouillon: @eric-chau, yes, the main issue comes from Twig: it catches every exeption thrown durng rendering and throw a new anonymous \Twig_Error_Exception.
So, @mickaelandrieu, rethrowing the previous exception if it's a FrontControllerException allows to get the real message and status code to be treated by ExceptionListener

crouillon:
ping @eric-chau, status?

eric-chau:
@crouillon sorry I thought this PR was already merged =) I think we can merge it cause we rethrow only if the previous exception is an instance of ``FrontControllerException``, which is exception that BackBee knows how to handle it right.

eric-chau:
Thank you @gobjila for your contribution! :+1:

PR #634, opened 2015-12-22
[SwiftMailer] fix encryption if not set in mailer.yml config file

grebniets:
Return `null` instead of `false` in case no encryption is detected in mailer.yml


eric-chau:
:+1:, nice catch.

PR #641, opened 2016-01-19
add title field in SEO form by default

flaviaF:


MetaData/MetaDataResolver.php
```diff
@@ -184,7 +184,14 @@ private function parseDefinition(array $definition, MetaData $metadata, Page $pa
             }
             $scheme = ('' === $metadata->getAttribute($attrname)) ? reset($attrvalue) : $metadata->getAttribute($attrname);
 
-            if (array_key_exists('default', $attrvalue)) {
+	    if (
+                    null !== $page
+                    && null !== $page->getTitle()
+                    && $metadata->getName() === 'title'
+		    && array_key_exists('default', $attrvalue)
+            ) {
```
eric-chau: Please take care about indentation and coding style =) Also you should test if $metadata is 'title' first so it can avoid you to do the check about `` null !== $page && null !== $page->getTitle()`` on every loop.

```php
if (
    'title' === $metadata->getName()
    && null !== $page
    && null !== $page->getTitle()
    && array_key_exists('default', $attrvalue)
) {
    $scheme = $page->getTitle();
}
```

crouillon:
:+1: Thank you @flaviaF

eric-chau:
thank you @flaviaF

PR #642, opened 2016-01-20
improved init of bundles in BBApplication when debug is setted to false

eric-chau:
We load every bundle's config only if the bundle loader is not restored. Tests on my side shows up an improvement of ~10 ms.


fkroockmann:
Nice :+1:

crouillon:
:+1: thanks @eric-chau

PR #643, opened 2016-01-20
[Renderer] improved Adapter\Twig and removed 'bbapplication.start' listener

eric-chau:
Optimized ``Adapter\Twig`` codes and added more configuration options for it. Removed strange `if statement`  that disable `Twig` cache on non CLI request.


eric-chau:
Created from my own fork.

PR #646, opened 2016-01-25
[Renderer] added TwigListener::onApplicationReady() to load twig extensions

eric-chau:
PR #645 initiated by myself removed the `TwigListener` and the loading of Twig's extensions is done in `Adapter\Twig::__construct()`. It can introduce bug in some cases like if the `Renderer` is invoked very early. So this specific behavior must be reverted.
Also I moved the listener to listen `bbapplication.init` instead of `bbapplication.start` so even is CLI mode Renderer will be configured as expected.


crouillon:
:+1:
thanks

PR #648, opened 2016-01-26
#531 redefine and develop search on pages

oanesbogdan:
[PAGE MANAGEMENT] redefine and develop search on pages 
- new parameter added to make the difference between page list and page tree
- translation for label "No page found." are needed for this languages: chines, russian, german


crouillon:
:+1:

Issue #649, opened 2016-01-27
[Rewriting][Bug] The URL of a page doesn't follow rewriting.yml rules


crouillon:
At the first valdation of the contents of a newly created page, the default scheme for URL rewriting is always applied, whatever specific rules defined

PR #650, opened 2016-01-27
[BUGFIX] #649 - Fix URL rewriting on first content vaidation + perf optimization

crouillon:


eric-chau:
:+1: thank you @crouillon, nice catch =)

Issue #655, opened 2016-02-02
[BUG] Some special characters truncate the serialized data in db


crouillon:
Some examples:
👩‍❤️‍👩 👨‍❤️‍👨 💏 👩‍❤️‍💋‍👩 👨‍❤️‍💋‍👨 👪 👨‍👩‍👧 👨‍👩‍

PR #656, opened 2016-02-02
[BUG] #655 - Some special characters truncate the serialized data in db

crouillon:


ClassContent/AbstractContent.php
```diff
@@ -284,9 +284,11 @@ public function __set($var, $value)
 
                 if (is_object($value) && $value instanceof AbstractClassContent) {
                     $value = $this->_addSubcontent($value);
+                } elseif ('scalar' === $type) {
+                    $val[] = array($type => json_encode($value, JSON_ERROR_NONE ));
+                } else {
+                    $val[] = array($type => $value);
                 }
```
eric-chau: This part introduces a code regression. $val[] = array($type => $value); is missing cause you move it into the `else` block

ClassContent/AbstractContent.php
```diff
@@ -284,9 +284,11 @@ public function __set($var, $value)
 
                 if (is_object($value) && $value instanceof AbstractClassContent) {
                     $value = $this->_addSubcontent($value);
+                } elseif ('scalar' === $type) {
+                    $val[] = array($type => json_encode($value, JSON_ERROR_NONE ));
```
eric-chau: Seems that JSON_ERROR_NONE is not a supported options. Valid ones: `JSON_HEX_QUOT, JSON_HEX_TAG, JSON_HEX_AMP, JSON_HEX_APOS, JSON_NUMERIC_CHECK, JSON_PRETTY_PRINT, JSON_UNESCAPED_SLASHES, JSON_FORCE_OBJECT, JSON_PRESERVE_ZERO_FRACTION, JSON_UNESCAPED_UNICOD`.

`JSON_ERROR_NONE` is one of the possible constant returned by `json_last_error()`.

ClassContent/AbstractContent.php
```diff
@@ -930,7 +932,10 @@ public function getData($var = null, $forceArray = false)
                 $value = end($values);
             }
 
-            if ($type !== 'scalar' && $type !== 'array') {
+            if ('scalar' === $type) {
+                $decode = json_decode($value);
+                $data[] = $decode ?: $value;
```
eric-chau: According to comment below, this is not a valid check.

crouillon:
Thanks @eric-chau but it's a little early to code review. I can't test locally for now so i alredy create the PR to get them
Renamed this PR [WIP] till it's ok

eric-chau:
ok @crouillon sorry!

eric-chau:
I will just add some explanations about how @crouillon fixed this issue. The issue is that for some special characters like emoji php `serialize()` is not able to do it correctly so it truncate the serialized string. This will result on an invalid insertion into database. To prevent this we first call `json_encode` on every scalar element then Doctrine can safely pass it to php `serialize()`. Indeed `json_serialize()` is able to convert special characters into valid character for php `serialize()`.

Hope this help you to understand the issue.

:+1:, thank you @crouillon

PR #657, opened 2016-02-02
Sequencer : avoid Doctrine cache and force reading from database

gobjila:
We need to get a refreshed new value from database in a foreach loop for exemple.
The Doctrine cache avoids that till the Entity is not refreshed.


crouillon:
:+1: Nice catch thank you

PR #662, opened 2016-02-03
[Renderer] fixed metadata helper use statement conflict

eric-chau:
`use statement` has conflict between `BackBee\DependencyInjection\Container` and `BackBee\Renderer\Helper\container` so this PR aims to fix it.

Below the error I got:
``Cannot use BackBee\\DependencyInjection\\Container as Container because the name is already in use in __ROOT_DIR__/vendor/backbee/backbee/Renderer/Helper/metadata.php on line 26``

### DIFF
--- a/Renderer/Helper/metadata.php
+++ b/Renderer/Helper/metadata.php
@@ -23,7 +23,6 @@
 
 namespace BackBee\Renderer\Helper;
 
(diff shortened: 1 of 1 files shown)


crouillon:
Nice catch, thank you  @eric-chau :+1:

PR #663, opened 2016-02-03
[Event] fixed major behavior bug in Dispatcher::triggerEvent()

eric-chau:
There is a wrong behavior in `BackBee\Event\Dispatcher::triggerEvent()` which dispatch event of classcontent parents first and the specific one as last, which is the opposite expected behavior.

Admit that you have a content called `Autoblock` that extends `ContentSet`.

Wrong behavior on dispatch on `render` event:

1. `classcontent.render`
2. `contentset.render`
3. `autoblock.render`

Right and expected behavior:

1. `autoblock.render`
2. `contentset.render`
3. `classcontent.render`

So this PR aims to fix it.

### DIFF
--- a/Event/Dispatcher.php
+++ b/Event/Dispatcher.php
@@ -23,13 +23,14 @@
 
 namespace BackBee\Event;
 
-use Symfony\Component\EventDispatcher\Event as sfEvent;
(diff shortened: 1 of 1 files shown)


crouillon:
:+1: very nice

PR #547, opened 2015-09-15
[REST] Fix some typos in getCollection of page

fkroockmann:
Fix some typos and use order by in each case.

PR #667, opened 2016-02-09
[DROPZONE] #547 - enable user to drop blocks only in zones that can handle the right render mode

crouillon:
* Introduce the exclusion of content type in accept arrays by prefixing classname with !
 * Allow to define accept array in subcontent defintions, for example:
```
ColumnDivider:
    properties:
        name : Column divider
        description: "Column divider"
        category: [Block]
    elements:
        first_container:
            type: BackBee\ClassContent\ContentSet
            accept: [BackBee\ClassContent\Home\HomeArticleContainer]
            maxentry: 1
```


eric-chau:
:+1:

crouillon:
Thanks @eric-chau

PR #668, opened 2016-02-09
Tree adjustments refs:#5362

oanesbogdan:
Changes for the following requests:
2. Pages that have subpages must be displayed before single pages
3. When creating a new page, display the page right after it parent . when reloading tree, page must go down after pages that have children
4. As soon as the toobar is loaded , load the tree in background task (for user to have it display quick when he'll click on it)(edited)


eric-chau:
Ok you added default value for order by with this PR. :+1:

oanesbogdan:
Yes this the only change for php code.
The rest of them are done in bb-core-je.

eric-chau:
okay, thank you @oanesbogdan :+1:

PR #669, opened 2016-02-09
[ENHANCEMENT] Allow to define the overriding configuration persist level

crouillon:
Introduce two optional configuration parameters to Config:config section:
 * _persist_per_context_: persist overriding config taking account of the context (true by default)
 * _persist_per_environment_: persist overriding config taking account of the environment (true by default)

### DIFF
--- a/Config/Persistor/File.php
+++ b/Config/Persistor/File.php
@@ -41,11 +41,27 @@ class File implements PersistorInterface
     private $app;
 
     /**
+     * Is a configuration is persisted by application context ?
+     *
+     * @var boolean
+     */
+    private $persistPerContext;
+
+    /**
+     * Is a configuration is persisted by application environment ?
+     *
+     * @var boolean
+     */
+    private $persistPerEnvironment;
+
+    /**
      * @see BackBee\Config\Persistor\PersistorInterface::__construct
      */
-    public function __construct(ApplicationInterface $app)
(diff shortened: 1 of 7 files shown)


Config/Persistor/File.php
```diff
@@ -75,12 +91,12 @@ public function persist(Config $config, array $configToPersist)
     private function getConfigDumpRightDirectory($baseDir)
     {
         $configDumpDir = $this->app->getRepository();
```
eric-chau: Shouldn't we use `::getBaseRepository()` instead? or just use `getRepository()` so you don't need to care anymore about environment?

eric-chau:
great PR @crouillon. Thank you :+1: just take a look at my comment ;)

Config/Persistor/File.php
crouillon: In fact, I think that we must take care of the context.

Till now the configuration persisting (File or Registry) was saved on the more precise level (ie taking care of context and environment).

I think it's always a good practice by default, but in some cases being able to save it on higher level can be useful ; example: the visitor website use a different environment than the contributor website but the bundle overriding config has to be shared => _persist_per_environment: false_ allows that now.

crouillon: At last, thanks @eric-chau you're right, getBaseRepsitory() has to be used

eric-chau: Cool then I will merge it when travis-ci is full green =)

PR #670, opened 2016-02-09
Optimize definitions call

fkroockmann:
Optimize time ~ 1s to ~150ms

### DIFF
--- a/Rest/Controller/ClassContentController.php
+++ b/Rest/Controller/ClassContentController.php
@@ -607,19 +607,29 @@ class ClassContentController extends AbstractRestController
      */
     private function getClassContentDefinitionsByCategory($name = null)
     {
-        $classnames = [];
-        if (null === $name) {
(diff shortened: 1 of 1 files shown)


eric-chau:
@fkroockmann can you explain me a bit more about this optimization please? I don't really get how it can drop from `1s` to `150ms` cause it seems that you just moved the code from `ClassContentManager` into `ClassContentController`.

eric-chau:
This move also mean that the result of `::getAllClassContentClassnames()` is not anymore in cache so we have to compute it on every call.

crouillon:
:+1:

fkroockmann:
With many classcontent like 300 for big project, the loop take min 600 ms.
I cache the call in db and the loop then

eric-chau:
Ok it's about the loop inside `ClassContentController::getClassContentDefinitionsByCategory()`. Nice catch, thank you :+1:

PR #671, opened 2016-02-09
Fix json encode on scalar element

fkroockmann:
Currently when scalar is empty or string the json decode return null instead of return the current string.

I add test to check if the json_decode dont have error and if the json decode return null.

I propose to set null if json decode throw error cause if we keep the current value, serialize will be broken

Let me know @eric-chau @crouillon

### DIFF
--- a/ClassContent/AbstractClassContent.php
+++ b/ClassContent/AbstractClassContent.php
@@ -408,8 +408,8 @@ abstract class AbstractClassContent extends AbstractContent
                     }
 
                     if ('scalar' === $type) {
-                        $decode = json_decode($value);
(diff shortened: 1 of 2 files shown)


eric-chau:
@fkroockmann your PR is very interesting. You're right, `json_decode` convert `null` and empty string into null so it's a good idea to leave with the initial value if `json_decode` succeed with no errors.
For your second suggestion about set the value to `null`, unfortunately we cannot do it cause we will break the compatibility with data stored in old database.

fkroockmann:
yes i have think about that, you are right, if the json decode throw an error we need to put the value to empty string, what do you think ?

fkroockmann:
I understand, i let the old value if the json decode throw an error

crouillon:
:+1: Thanks @fkroockmann

PR #672, opened 2016-02-10
[Renderer] fixed metadata helper cause it throws warning on some use case

eric-chau:
Currently, `Renderer/Helper/metadata::__invoke` throws warning when `$metadata` is `null`.


crouillon:
:+1: Thanks @eric-chau

PR #675, opened 2016-02-11
[PAGE MANAGEMENT] refs:520 delete page and delete trash is KO

crouillon:
* Introduce new REST action in PageController: deleteCollectionAction() to handle multi-delete request
 * Refactoring of softDelete (ie only state change to trash)
 * Fix hardDelete (ie removing from the db)

### DIFF
--- a/ClassContent/Repository/ClassContentRepository.php
+++ b/ClassContent/Repository/ClassContentRepository.php
@@ -893,6 +893,13 @@ class ClassContentRepository extends EntityRepository
      */
     public function deleteContent(AbstractClassContent $content, $mainContent = true)
     {
+        if (!$this->getEntityManager()->contains($content)) {
+            // Ensure $content is managed
+            $content = $this->getEntityManager()
+                    ->getRepository(AbstractClassContent::getFullClassname($content))
+                    ->find($content->getUid());
+        }
+
         $parents = $this->getParentContents($content);
         $media = $this->_em->getRepository('BackBee\NestedNode\Media')->findOneBy([
             '_content' => $content->getUid(),
@@ -929,7 +936,7 @@ class ClassContentRepository extends EntityRepository
             }
 
             $this->cleanUpContentHardDelete($content);
-            $this->_em->remove($content);
+            $this->_em->remove($content->setMainNode(null));
         }
     }
 
--- a/Rest/Controller/PageController.php
+++ b/Rest/Controller/PageController.php
@@ -518,10 +518,49 @@ class PageController extends AbstractRestController
         }
     }
 
+    /**
+     * Moves the pages in trash.
+     * 
+     * @param  Page $page The page to be moved in trash.
+     * 
+     * @throws BadRequestHttpException Occures if $page is a root.
+     */
+    private function softDelete(Page $page)
+    {
+        if ($page->isRoot()) {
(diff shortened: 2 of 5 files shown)


ClassContent/Repository/ClassContentRepository.php
```diff
@@ -929,7 +936,7 @@ public function deleteContent(AbstractClassContent $content, $mainContent = true
             }
 
             $this->cleanUpContentHardDelete($content);
-            $this->_em->remove($content);
+            $this->_em->remove($content->setMainNode(null));
```
eric-chau: Nice catch =)

Rest/Controller/PageController.php
```diff
@@ -518,10 +518,49 @@ private function updatePageState(Page $page, $state)
...
+    /**
+     * Moves the pages in trash.
+     * 
+     * @param  Page $page The page to be moved in trash.
+     * 
+     * @throws BadRequestHttpException Occures if $page is a root.
+     */
+    private function softDelete(Page $page)
+    {
+        if (true === $page->isRoot()) {
```
eric-chau: you don't need the `true ===` part, `if ($page->isRoot()) {` is enough clear as it returns a `boolean`.

Rest/Controller/PageController.php
```diff
@@ -518,10 +518,49 @@ private function updatePageState(Page $page, $state)
...
+            throw new BadRequestHttpException('Cannot remove root page of a site.');
+        }
+
+        $this->granted('DELETE', $page);
+        $this->granted('EDIT', $page->getParent()); // user must have edit permission on parent
+
+        if (true === $page->isOnline(true)) {
```
eric-chau: you don't need the `true ===` part, `if ($page->isOnline(true)) {`.

Rest/Controller/PageController.php
```diff
@@ -579,23 +618,75 @@ public function patchAction(Page $page, Request $request)
...
-        if (true === $page->isRoot()) {
-            throw new BadRequestHttpException('Cannot remove root page of a site.');
+        if ($page->isDeleted()) {
+            $this->hardDelete($page);
+        } else {
+            $this->softDelete($page);
         }
 
-        $this->granted('DELETE', $page);
+        $this->getEntityManager()->flush($page);
```
eric-chau: Would it be better to call `flush()` without argument instead of `$page`? To be sure that every entity like section or content are deleted too. Let me know maybe it's already configured to be removed in cascade.

Rest/Controller/PageController.php
```diff
@@ -579,23 +618,75 @@ public function patchAction(Page $page, Request $request)
...
+            throw new BadRequestHttpException('uid is missing.');
         }
 
-        $this->getPageRepository()->toTrash($page);
-        
-        $this->getEntityManager()->flush($page);
+        $result = [];
+        $statusCode = 204;
+        $pages = $this->getPageRepository()->findBy(['_uid' => $uids]);
+        foreach($pages as $page) {
```
eric-chau: missing a space between `foreach` and `(` =)

eric-chau:
Good job @crouillon, thank you :+1: =)

crouillon:
Thanks @eric-chau

PR #680, opened 2016-02-18
Enabling to find if a media is linked to a content

hbaptiste:
Enabling to find if a media is linked to a content

### DIFF
--- a/NestedNode/Repository/MediaRepository.php
+++ b/NestedNode/Repository/MediaRepository.php
@@ -56,6 +56,17 @@ class MediaRepository extends EntityRepository
             ])
         ;
 
+        /* find media by content */
+        $contentType = isset($cond['contentType']) ? $cond['contentType'] : null;
+        if (null !== $contentType) {
+            $contentUid = isset($cond['contentUid']) ? $cond['contentUid'] : null;
+            $content = $this->_em->find($contentType, $contentUid);
+            if (null !== $content) {
+                $parents = $content->getParentContent();
+                $q->andWhere('m._content IN (:contents)')->setParameter('contents', $parents->toArray());
+            }
+        }
+
         $typeField = isset($cond['typeField']) && 'all' !== $cond['typeField'] ? $cond['typeField'] : null;
         if (null !== $typeField) {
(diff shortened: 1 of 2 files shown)


NestedNode/Repository/MediaRepository.php
```diff
@@ -56,6 +56,18 @@ public function getMedias(MediaFolder $mediafolder, $cond, $order_sort = '_title
             ])
         ;
 
+        /* find media by content */
+        $contentType = isset($cond['contentType']) ? $cond['contentType'] : null;
+        if (null !== $contentType) {
+            $contentUid = isset($cond['contentUid']) ? $cond['contentUid'] : null;
+            $classname = \BackBee\ClassContent\AbstractContent::getClassnameByContentType($contentType);
```
eric-chau: Can you add `use BackBee\ClassContent\AbstractContent` statement so you can write it like that `AbstractContent::getClassnameByContentType($contentType);` please?

Rest/Controller/MediaController.php
```diff
@@ -156,7 +158,13 @@ public function postAction(Request $request)
...
 
         $content = $this->getClassContentManager()->findOneByTypeAndUid($contentType, $contentUid);
-        $mediaFolder = $this->getMediaFolderRepository()->find($mediaFolderUid);
+
+        if (null === $mediaFolderUid) {
+            $mediaFolder = $this->getMediaFolderRepository()->getRoot();
+        } else {
+            $mediaFolder = $this->getMediaFolderRepository()->find($mediaFolderUid);
+        }
+
```
eric-chau: What if `$mediaFolder` is equal to `null`?

NestedNode/Repository/MediaRepository.php
eric-chau: you should also surround this call by a try catch cause this method can throw `BackBee\ClassContent\Exception\InvalidContentTypeException` if the provided content type is invalid.

eric-chau:
:+1:, thank you @hbaptiste, phease take time to work on my comments.

PR #681, opened 2016-02-18
Fix getAccept() on old version layouts

crouillon:


ClassContent/AbstractContent.php
```diff
@@ -1208,11 +1208,11 @@ public function jsonSerialize($format = self::JSON_DEFAULT_FORMAT)
             'modified'   => $this->_modified->getTimestamp(),
             'revision'   => $this->_revision,
             'parameters' => $this->getAllParams(),
-            'accept'     => array_map(
+            'accept'     =>  array_map(
```
eric-chau: There is an unnecessary blank space before `array_map()`

ClassContent/AbstractContent.php
```diff
@@ -1290,54 +1290,90 @@ private function formatJsonData(array $data, $format)
...
+            return  array_map(['BackBee\ClassContent\AbstractContent', 'getFullClassname'], $classname);
         }
 
+        $isExcluded = self::isExcludedClassname($classname);
+
         try {
-            if (0 !== strpos($classname, self::CLASSCONTENT_BASE_NAMESPACE)) {
-                $classname = self::CLASSCONTENT_BASE_NAMESPACE.$classname;
-            }
             class_exists($classname);
```
eric-chau: I don't get how it can work cause you didn't add the `self::CLASSCONTENT_BASE_NAMESPACE` prefix if $classname is a string. Am I wrong? Also, in case `$classname` is an instance of `AbstractClassContent` I think it won't work.

ClassContent/AbstractContent.php
```diff
@@ -1290,54 +1290,90 @@ private function formatJsonData(array $data, $format)
...
+     */
+    private static function isExcludedClassname(&$classname)
+    {
+        $isExcluded = false;
+
+        if (is_object($classname)) {
+            if (!$classname instanceof AbstractClassContent) {
+                throw new \InvalidArgumentException('First parameter must be a string or an AbstractClassContent object.');
+            }
+            $classname = ClassUtils::getRealClass($classname);
```
eric-chau: I think this is a bad design cause we change the variable type from `object` to `string` sometimes. It takes many time for @fkroockmann and me to understand how it works.

ClassContent/AbstractContent.php
crouillon: At this point $classname is always a string
(call by ref in isExcludedClassname, see note below)

ClassContent/AbstractContent.php
crouillon: getFullClassname and getShortClassname return always a string but can accept by convenience a mixed parameter but, at last, a string with fully namspaced classname is treated: 
 * an array : recursive call to itself
 * an object => getRealClass() is called
 * a string : test if it's a valid classname

eric-chau: Ok. But what I wanted to mean is that it's not a good idea to change the type of a variable especially if it is passed by reference (i.e. `$classname`). It is not obvious and logic to guess that passing an object by reference will convert it into string.

crouillon: well it's a private

eric-chau:
:+1: thank you @crouillon =)

PR #682, opened 2016-02-22
Add main node state for content

fkroockmann:
Add main node state to know if the content have them page online or not

### DIFF
--- a/ClassContent/AbstractContent.php
+++ b/ClassContent/AbstractContent.php
@@ -1199,6 +1199,8 @@ abstract class AbstractContent implements ObjectIdentifiableInterface, Renderabl
      */
     public function jsonSerialize($format = self::JSON_DEFAULT_FORMAT)
     {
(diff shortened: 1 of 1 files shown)


ClassContent/AbstractContent.php
```diff
@@ -1219,6 +1221,7 @@ function ($classname) {
             'elements'   => $this->computeElementsToJson($this->getData()),
             'has_elements' => count($this->getData()) === 0 ? false : true,
             'extra'      => [],
+            'main_node_state' => (null !== $node) ? (int) $node->isOnline() : null,
```
eric-chau: It's clearer to name the parameter like this: `'is_mainnode_online' => $node ? $node->isOnline() : null` doing it like this will prevent you to explain the meaning of `1` or `2` etc.

eric-chau:
:+1: good job @fkroockmann

PR #683, opened 2016-02-25
Merge options with accept on contentset

fkroockmann:
Options is used when value is not set on the YML.


eric-chau:
:+1: thank you =)

PR #684, opened 2016-03-01
Add return this on ContentSet::setParam function

fkroockmann:
The function chain is currently not available on Contenset


ClassContent/ContentSet.php
```diff
@@ -614,6 +614,8 @@ public function setParam($key, $value = null)
                 $this->setAccept($value);
             }
         }
+
+        return $this->getContentInstance();
```
crouillon: return $this; is preferable
$this->getContentInstance() has to be used in AbstractContent because it is generalized by both AbstractClassContent and Revision objects in order to return always a content instance and never a revision instance

PR #685, opened 2016-03-02
[ClassContent] fixed AbstractContent which has not main node attribute

eric-chau:
Fixed `AbstractContent` has not main node attribute but `AbstractClassContent` does.


fkroockmann:
:+1:

PR #689, opened 2016-03-11
Removed array_columun function - not suported by PHP 5.4

oanesbogdan:
- "array_column" is suported only by PHP 5 >= 5.5.0
- made both options mandatory "site_label" & "site_domain"


crouillon:
:+1: Thanks @oanesbogdan

PR #692, opened 2016-03-16
Fix accept parameters creation for contentset

fkroockmann:
For accept parameters, we have two ways:

````
accept: ['MyContent', 'MyContent2']     =>   $options  

accept:      => $options['value']
    type: ContentTypeSelector
    value: ['MyContent', 'MyContent2']
````

### DIFF
--- a/ClassContent/ContentSet.php
+++ b/ClassContent/ContentSet.php
@@ -654,7 +654,10 @@ class ContentSet extends AbstractClassContent implements \Iterator, \Countable
 
         if ('accept' === $var) {
             if (null !== $options) {
-                $this->_addAcceptedType(array_merge((array) $options, $this->getParamValue('accept')));
+
+                $accept = array_key_exists('value', $options) ? $options['value'] : $options;
+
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by crouillon** [comment]
:+1:, thanks @fkroockmann

File: ClassContent/ContentSet.php [hunk 1]
```diff
@@ -654,7 +654,10 @@ protected function defineParam($var, $options = null)
 
         if ('accept' === $var) {
             if (null !== $options) {
-                $this->_addAcceptedType(array_merge((array) $options, $this->getParamValue('accept')));
+
+                $accept = isset($options['value']) ? $options['value'] : $options;
```
eric-chau: Use `array_key_exists` instead of `isset` cause if you have this:

```
parameters:
    accept:
        type: ContentTypeSelector
        value: ~
```
The resulted accept parameter will be wrong.

eric-chau:
:+1: nice catch @fkroockmann

ClassContent/ContentSet.php
fkroockmann: that cannot be a null value, in parameters you must declare an array

PR #693, opened 2016-03-17
[BUGFIX] Ensure PostResponseEvent is triggered with a Response

crouillon:
For compatibility consideration, a WARNING level log is generated if a controller does not return an object Response and a new empty Response is created.

### DIFF
--- a/Controller/FrontController.php
+++ b/Controller/FrontController.php
@@ -565,6 +565,16 @@ class FrontController implements HttpKernelInterface
             $actionArguments = $controllerResolver->getArguments($request, $controller);
 
             $response = call_user_func_array($controller, $actionArguments);
+            if (!($response instanceof Response)) {
+                $this->application->warning(sprintf(
+                    'Action %s from controller %s should return an object Response, got %s.',
+                    $request->attributes->get('_action'),
+                    get_class($controller[0]),
+                    gettype($response)
+                ));
+
+                $response = new Response();
+            }
             $dispatcher->dispatch($eventName.'.postcall', new Event\PostResponseEvent($response, $request));
 
             return $response;


Controller/FrontController.php
```diff
@@ -565,7 +565,7 @@ protected function invokeAction($type = self::MASTER_REQUEST)
             $actionArguments = $controllerResolver->getArguments($request, $controller);
 
             $response = call_user_func_array($controller, $actionArguments);
-            $dispatcher->dispatch($eventName.'.postcall', new Event\PostResponseEvent($response, $request));
+            $dispatcher->dispatch($eventName.'.postcall', new Event\PostResponseEvent($response ?: new Response(), $request));
```
eric-chau: Shouldn't we do this? `$response = call_user_func_array($controller, $actionArguments) ?: new Response();` cause if $response is modified during the dispatch of `PostResponseEvent` and $response is null it won't work as expected.

crouillon: hum, I think it's the same: $response is tested before the PostResponseEvent is dispatched

eric-chau: No cause in your case, if `$response` is equal to `null`:

```php
$response = call_user_func_array($controller, $actionArguments); // $response = null
$dispatcher->dispatch($eventName.'.postcall', new Event\PostResponseEvent($response ?: new Response(), $request));

return $response; // $response still equal to null
```

in `$response ?: new Response()`, if `$response` is equal to null, `new Response()` is passed to `PostResponseEvent::__construct` as anonymous variable.

crouillon: you're right, i didn't notice the return $response; :-(
Thanks @eric-chau

Controller/FrontController.php
```diff
@@ -565,6 +565,15 @@ protected function invokeAction($type = self::MASTER_REQUEST)
...
 
             $response = call_user_func_array($controller, $actionArguments);
+            if (!($response instanceof Response)) {
+                $this->application->warning(sprintf(
+                        'Action %s from controller %s should return an object Response, got %s.',
+                        $request->attributes->get('_action'),
+                        get_class($controller[0]),
+                        gettype($response)
+                ));
+                $response = new Response();
```
eric-chau: Should'nt we test if `$response` is a string? If it's the case we could pass it to `Response` constructor like this: `$response = new Response(is_string($reponse) ? $response : '');`
It's current to return the result of `Renderer::render()` especially for Symfony users.

Controller/FrontController.php
```diff
@@ -565,6 +565,15 @@ protected function invokeAction($type = self::MASTER_REQUEST)
...
+            }
```
eric-chau: Can you add a new line please?

Controller/FrontController.php
```diff
@@ -565,6 +565,15 @@ protected function invokeAction($type = self::MASTER_REQUEST)
             $actionArguments = $controllerResolver->getArguments($request, $controller);
 
             $response = call_user_func_array($controller, $actionArguments);
+            if (!($response instanceof Response)) {
+                $this->application->warning(sprintf(
+                        'Action %s from controller %s should return an object Response, got %s.',
```
eric-chau: there is an extra tab for all these lines no?

eric-chau:
:+1:

Controller/FrontController.php
crouillon: I'm not so confortable with that idea: Symfony doc explicitly say that the goal of a controller is always the same: create and return a Response object.

eric-chau: Ok =)

crouillon:
Thx @eric-chau

PR #694, opened 2016-03-23
[Security] updated its services to remove duplicated role voter

eric-chau:
Removed duplicated `@security.voter.bb_role` in `security.access.decision_manager` service declaration.


fkroockmann:
:+1:

crouillon:
:+1:

PR #695, opened 2016-03-23
removed use of datacontent helper which does not exist anymore

eric-chau:
Still have some phtml views that have reference to datacontent helper but it does not exist anymore. So this PR aims to fix it.


fkroockmann:
:+1:

PR #696, opened 2016-03-29
Console command (bbapp:create_site) compatibility improvements

gobjila:
Hello, I improved some "exit codes" and "console exceptions" in order to make the command compatible with calls from another console commands or from a web controller.

### DIFF
--- a/Console/Command/CreateWebsiteCommand.php
+++ b/Console/Command/CreateWebsiteCommand.php
@@ -19,6 +19,7 @@
  * along with BackBee. If not, see <http://www.gnu.org/licenses/>.
  *
  * @author Bogdan Oanes <bogdan.oanes@lp-digital.fr>
+ * @author Alexandre Gobjila <alexandre.gobjila@lp-digital.fr>
  */
 
 namespace BackBee\Console\Command;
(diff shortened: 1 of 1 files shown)


Console/Command/CreateWebsiteCommand.php
```diff
@@ -104,12 +105,17 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+        try{
+            $this
+                ->init($input, $output)
+                ->runSitesYmlProcess()
+                ->runLayoutsGenerationProcess()
+            ;
+        } catch (\Exception $e) {
+            $output->writeln('<error>'.$e->getMessage().'</error>');
+            return 1;
+        }
```
crouillon: Could you explicitly add a return 0; when everything ok ?

Console/Command/CreateWebsiteCommand.php
```diff
@@ -104,12 +105,17 @@ protected function execute(InputInterface $input, OutputInterface $output)
             throw new \InvalidArgumentException('Both options `site_label` and `site_domain` are required.');
         }
 
-    	$this
-    		->init($input, $output)
-    		->runSitesYmlProcess()
-    		->runLayoutsGenerationProcess()
-		;
-    }        
+        try{
```
eric-chau: take care about coding style please, a whitespace is missing between `try` and `{`

Console/Command/CreateWebsiteCommand.php
eric-chau: agree with @crouillon

crouillon:
Thanks @gobjila

PR #699, opened 2016-04-07
[Resources] removed and renamed classcontent element phtml templates

eric-chau:
Renamed classcontent element phtml templates to match with its YAML equivalent files.

ping @fkroockmann, @crouillon


crouillon:
:+1:

PR #701, opened 2016-04-08
[Bundle] updated BundleLoader::buildBundleDefinition() to also call ::started()

eric-chau:
The first time a bundle is getting from the container, the method `::start()` is called. This is the expected behavior but we also have to invoke `::started()` method otherwise `::isStarted()` will always return `false`.

ping @crouillon


crouillon:
:+1:

PR #702, opened 2016-04-08
[Controller] updated FrontController to return response if exists after dispatch of  event

eric-chau:
Currently, even if the dispatch of `kernel.request` event find a response, we don't take it in account and we keep running like it does not happen.

ping @crouillon


crouillon:
Yes !
Thx @eric-chau 
:+1:

PR #703, opened 2016-04-13
Allow to define a specific vendor directory in MockBBApplication

crouillon:
Required for example to test bundles: the vendor dir is not in BackBee dir.


eric-chau:
:+1:

PR #706, opened 2016-04-21
[ClassContent] updated RevisionRepository to handle corrupted drafts

eric-chau:
Surrounded `RevisionRepository::getAllDrafts` with `try-catch` to handle corrupted drafts.


crouillon:
Nice catch, thx @eric-chau :+1:

PR #707, opened 2016-04-27
689 - Check if link is created between page and content

flaviaF:
https://github.com/backbee/backbee-standard/issues/689
Set the main node property to content drag & dropped in a main column of a page.


crouillon:
:+1: Thank you @flaviaF

PR #710, opened 2016-05-02
Fix CleanOrphan command to enable short classnames support

crouillon:
Since short classnames are used for contents, the command `content:clean_orphan` must be updated to support them.


eric-chau:
:+1: thanks @crouillon

PR #711, opened 2016-05-02
Add registered method to handle fatal errors

crouillon:
Allow to write in log file the last fatal error catched, especially database connection errors.

### DIFF
--- a/NestedNode/Repository/MediaRepository.php
+++ b/NestedNode/Repository/MediaRepository.php
@@ -23,6 +23,7 @@
 
 namespace BackBee\NestedNode\Repository;
 
+use BackBee\ClassContent\AbstractClassContent;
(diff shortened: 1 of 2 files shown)


BBApplication.php
```diff
@@ -112,6 +113,8 @@ class BBApplication implements ApplicationInterface, DumpableServiceInterface, D
      */
     public function __construct($context = null, $environment = null, $overwrite_config = false)
     {
+        register_shutdown_function([$this, 'onFatalError']);
```
eric-chau: You should put this instruction after `::initContainer` cause `BBApplication::__call` relies on BackBee's service container.

eric-chau:
:+1: nice addition @crouillon, thank you

PR #713, opened 2016-05-02
Add registered method to handle fatal errors

crouillon:
Allow to write in log file the last fatal error catched, especially database connection errors.


eric-chau:
:+1: nice addition, thanks!

PR #714, opened 2016-05-03
[BugFix] Fix IndexationRepository::getNodeUids()

crouillon:
Avoid a 'Array to string conversion' notice while invoking the method.


eric-chau:
:+1:

PR #715, opened 2016-05-04
updated ::commit() and ::revert() of ClassContentManager

eric-chau:
Updated `::commit()` and `::revert()` so the second argument is optional. if it is not provided, the content is fully commited of reverted.


crouillon:
:+1:  thx @eric-chau

PR #717, opened 2016-05-16
[Config] fixed Configurator::doBundleConfigExtend() to apply registry override only if save_in_registry exists and equals to true

eric-chau:
The if condition in `Configurator::doBundleConfigExtend()` to determine if we should apply registry override or not is wrong. Currently even if `save_in_registry` is equal to `false` Configurator still run the override by registry process. This PR aims to fix it.


crouillon:
Nice catch @eric-chau , thx :+1:

PR #718, opened 2016-05-17
[Installer] optimized EntityFinder::getNamespace()

eric-chau:
Optimized `EntityFinder::getNamespace()` to reduce the string to tokenize by `token_get_all()`. In some cases the memory size limit is reached.


crouillon:
:+1:

PR #719, opened 2016-05-17
[Util] updated the way EntityManagerCreator sets Doctrine metadata cache

eric-chau:
Updated the way `EntityManagerCreator` sets Doctrine metadata cache according to provided configuration. In some case the service identifier is replaced by the object on `Config` compilation. To prevent the fatal error due to it, I added additionnal checks to do the right thing on string and on object.

ping @crouillon


crouillon:
:+1: Thx @eric-chau

PR #720, opened 2016-05-18
Introduce Rest\RevisionController + Add ClassContentManager::revertToRevision() feature

crouillon:
Introduce three new REST services:
 * `RevisionController::getAction()`: returns a specific revision by uid.
 * `RevisionController::getCollectionByContentAction()`: Returns collection of revisions for a content.
 * `ClassContentController::patchDraftAction()`: Reverts user's draft of a content to a specific revision number. Possible values for revision number are:
     *  negative value: revert to current revision decrease from value.
     *  empty value: revert to last committed revision.
     *  positive value: revert to specific revision if exists.


eric-chau:
:+1: good job @crouillon

PR #721, opened 2016-05-18
[REST] KeywordController::getCollectionAction() accepts now uids parameter

crouillon:
/rest/2:/keyword?uids=uid1,uid2,.. is now accepted by the service


eric-chau:
:+1:

PR #722, opened 2016-05-18
[Installer] fixed EntityFinder::getNamespace() to refer to ';' instead of '{'

eric-chau:
Changed `EntityFinder::getNamespace()` to refer to the first occurence of `;` instead of `{` cause in some cases we broke comment block so `token_get_all()` generates PHP warnings.


crouillon:
:+1:

PR #723, opened 2016-05-20
[ClassContent] updated ClassContentQueryBuilder::addClassFilter 

eric-chau:
Updated `ClassContentQueryBuilder::addClassFilter` to ensure that requested classes are loaded cause sometimes we get error about classcontent that are not part of AbstractClassContent discriminators.

ping @crouillon


crouillon:
Very nice @eric-chau :+1:

PR #729, opened 2016-05-31
[ClassContent] updated RevisionRepository::getAllDrafts to delete revision of unknown content

eric-chau:
This issue may appear if you upgrade BackBee's version and especially when you didn't import every classcontent type. This PR aims to avoid runtime exception by deleting these revisions that belong to old content.


crouillon:
:+1: Thx @eric-chau

PR #730, opened 2016-06-07
[ClassContent] improved FileJsonSerializeTrait::jsonSerialize()

eric-chau:
Added more checks to `FileJsonSerializeTrait::jsonSerialize()` to prevent error like this:

`PHP Warning: json_decode() expects parameter 1 to be string, array given in vendor/backbee/backbee/ClassContent/Traits/Element/FileJsonSerializeTrait.php on line 22`


crouillon:
:+1: thx @eric-chau

PR #732, opened 2016-06-10
730 - tree is empty after deleting a page with sub-pages

flaviaF:
https://github.com/backbee/backbee-standard/issues/730

### DIFF
--- a/Event/Listener/PageListener.php
+++ b/Event/Listener/PageListener.php
@@ -92,6 +92,7 @@ class PageListener
         $updateParents = false;
         $new = $old = null;
 
+        $pageCountModifier = -1;
         if ($eventArgs instanceof PreUpdateEventArgs) {
             if ($page instanceof Page && $eventArgs->hasChangedField('_section')) {
                 $old = $eventArgs->getOldValue('_section');
@@ -107,6 +108,9 @@ class PageListener
                 if ($page->getParent() !== null) {
                     if ($eventArgs->getNewValue('_state') >= 4) {
                         $old = $page->getParent()->getSection();
+                        if (!$page->isLeaf()) {
+                            $pageCountModifier = 0;
+                        }
                     } else {
                         $new = $page->getParent()->getSection();
                     }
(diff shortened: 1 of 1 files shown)


NestedNode/Repository/PageRepository.php
```diff
@@ -807,6 +807,8 @@ public function toTrash(Page $page)
                     ->getRepository('BackBee\NestedNode\Section')
                     ->createQueryBuilder('n')
                     ->select('n._uid')
+                    ->where('n._uid != :curentPage')
+                    ->setParameter('curentPage', $page->getUid())
                     ->andIsDescendantOf($page->getSection());
```
crouillon: hum, we're looking for Section here and you exclude a Page, could you rather try `->andIsDescendantOf($page->getSection(), true)` that should exclude the current section, what we need if I understand

NestedNode/Tests/Repository/PageRepositoryTest.php
```diff
@@ -748,13 +748,9 @@ public function testToTrash()
         $page2 = $this->repository->find('page2');
         $page3 = $this->repository->find('page3');
 
-        $this->assertEquals($page3, $this->repository->toTrash($page3));
-        $this->assertEquals(Page::STATE_DELETED, $page3->getState());
```
crouillon: You can't do that, if $page3 is put in the trash then its state is deleted

NestedNode/Tests/Repository/PageRepositoryTest.php
```diff
@@ -748,13 +748,9 @@ public function testToTrash()
...
-
         $this->assertEquals($section1, $this->repository->toTrash($section1));
         self::$em->refresh($page1);
         self::$em->refresh($page2);
-        $this->assertEquals(Page::STATE_DELETED, $section1->getState());
```
crouillon: Same thing with $section1, so I think that we musn't exclude the current section in toTrash method()

crouillon:
Thx @flaviaF

gobjila:
Thanks @flaviaF

PR #733, opened 2016-06-10
[BUGFIX] An existing group was looked for by its description

crouillon:
The correct field search is the name of the group.
Also take account of optional description


eric-chau:
:+1:

PR #735, opened 2016-06-20
fix delete page with section

gobjila:
Fix recursive delete for a page with sub-sections
Thanks @crouillon for your help.


crouillon:
:+1: thanks @gobjila

PR #743, opened 2016-07-04
[Bundle] updated BundleLoader::loadFullBundles() to load every bundle services first

eric-chau:
Updated `BundleLoader::loadFullBundles()` to load every bundle's services first. Actually, we load every bundle completely one by one so `::loadRoutes()` is invoked too early. It means that only the first declared bundle can override `routing` service definition. Which is not what we want. This PR aims to fix it.


crouillon:
:+1: thx @eric-chau

PR #746, opened 2016-07-25
fix authentication method to return a boolean.

miladb:
Fix this error : no authenticationprovider found for token of class usernamepasswordtoken

### DIFF
--- a/Security/Authentication/Provider/UserAuthenticationProvider.php
+++ b/Security/Authentication/Provider/UserAuthenticationProvider.php
@@ -149,16 +149,15 @@ class UserAuthenticationProvider implements AuthenticationProviderInterface
      */
     private function authenticateWithEncoder(TokenInterface $token, UserInterface $user)
     {
-        try {
-            $classname = \Symfony\Component\Security\Core\Util\ClassUtils::getRealClass($user);
-            if (true === $this->_encoderFactory
-                            ->getEncoder($classname)
-                            ->isPasswordValid($user->getPassword(), $token->getCredentials(), $user->getSalt())) {
-                return new UsernamePasswordToken($user, $user->getPassword(), $user->getRoles());
-            }
-        } catch (Exception $e) {
-            return false;
+        $classname = \Symfony\Component\Security\Core\Util\ClassUtils::getRealClass($user);
+
+        if (true === $this->_encoderFactory
+                ->getEncoder($classname)
+                ->isPasswordValid($user->getPassword(), $token->getCredentials(), $user->getSalt())) {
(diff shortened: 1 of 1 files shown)


Security/Authentication/Provider/UserAuthenticationProvider.php
```diff
@@ -149,16 +149,15 @@ private function authenticateUser(TokenInterface $token, UserInterface $user)
...
-            }
-        } catch (Exception $e) {
-            return false;
+        $classname = \Symfony\Component\Security\Core\Util\ClassUtils::getRealClass($user);
+
+        if (true === $this->_encoderFactory
+                ->getEncoder($classname)
+                ->isPasswordValid($user->getPassword(), $token->getCredentials(), $user->getSalt())) {
+            return new UsernamePasswordToken($user, $user->getPassword(), $user->getRoles());
         }
```
eric-chau: Hey @miladb, thank you for your contribution! You should leave the `try-catch` and move the `return false;` out of the `catch` block. Can you do it please?

miladb: Hey @eric-chau. I can do it but in the catch, you want to throw the exception or just return false ?

eric-chau:
@miladb, you don't have to do anything in the `catch` block cause we execute the same code in both case, returning `false`.

Security/Authentication/Provider/UserAuthenticationProvider.php
eric-chau: @miladb, you don't have to do anything in the `catch` block cause we execute the same code in both case, returning `false`.

Security/Authentication/Provider/UserAuthenticationProvider.php
```diff
@@ -152,13 +152,15 @@ private function authenticateWithEncoder(TokenInterface $token, UserInterface $u
...
             $classname = \Symfony\Component\Security\Core\Util\ClassUtils::getRealClass($user);
             if (true === $this->_encoderFactory
-                            ->getEncoder($classname)
-                            ->isPasswordValid($user->getPassword(), $token->getCredentials(), $user->getSalt())) {
+                    ->getEncoder($classname)
+                    ->isPasswordValid($user->getPassword(), $token->getCredentials(), $user->getSalt())) {
                 return new UsernamePasswordToken($user, $user->getPassword(), $user->getRoles());
             }
         } catch (Exception $e) {
             return false;
```
eric-chau: Sorry I might be not enough clear, but can you remove this line please? ^^ So everything is fine.

eric-chau:
thank you @miladb!

PR #747, opened 2016-07-28
Add Index on type column of registry table

Jeremy95:


eric-chau:
thank you @Jeremy95 :+1:

PR #748, opened 2016-07-29
add unique index for each of three column instead one index for them

Jeremy95:
before (with one index on here 2 column key and scope)
avant

after (with one index for each of three column scope, key, type)
apres


eric-chau:
Ok thank you @Jeremy95 :+1:

PR #749, opened 2016-07-29
add unique index for each of four column instead one index for all of…

Jeremy95:
… them

same as registry. 

significant gain performance.

### DIFF
--- a/NestedNode/Section.php
+++ b/NestedNode/Section.php
@@ -39,8 +39,6 @@ use BackBee\Site\Site;
  * @ORM\Entity(repositoryClass="BackBee\NestedNode\Repository\SectionRepository")
  * @ORM\Table(name="section",
  *    indexes={
- *              @ORM\Index(name="IDX_UID", columns={"uid"}),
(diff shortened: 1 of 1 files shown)


NestedNode/Section.php
```diff
@@ -37,7 +37,14 @@
  * @author      c.rouillon <charles.rouillon@lp-digital.fr>
  *
  * @ORM\Entity(repositoryClass="BackBee\NestedNode\Repository\SectionRepository")
- * @ORM\Table(name="section",indexes={@ORM\Index(name="IDX_TREE_SECTION", columns={"uid", "root_uid", "leftnode", "rightnode"})})
+ * @ORM\Table(name="section",
+ *    indexes={
+ *              @ORM\Index(name="IDX_UID", columns={"uid"}),
+ *              @ORM\Index(name="IDX_ROOT_UID", columns={"root_uid"}),
```
eric-chau: There is already an index for primary key and foreign key, we don't have to add another index for them.

eric-chau:
:+1: thank you @Jeremy95

PR #750, opened 2016-08-01
Sort elements by uids preserve order

cmarianhodis:
Preserve the order of the elements

### DIFF
--- a/Rest/Controller/AbstractRestController.php
+++ b/Rest/Controller/AbstractRestController.php
@@ -250,4 +250,33 @@ abstract class AbstractRestController extends Controller implements RestControll
     {
         return $this->getRequest()->attributes->get($name);
     }
+
+    /**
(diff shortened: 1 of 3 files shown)


Rest/Controller/AbstractRestController.php
```diff
@@ -250,4 +250,35 @@ protected function getEntityFromAttributes($name)
...
+     * @param array $uids
+     * @param Paginator $contents
+     *
+     * @return null|Paginator
+     */
+    protected function sortByUids($uids, $contents)
+    {
+        if (false === is_array($uids)) {
+            $uids = array($uids);
+        }
```
eric-chau: You can avoid this `if` condition like this: `$uids = (array) $uids;` It will cast `$uids` into array if it is not, else it does nothing.

eric-chau:
Nice job @cmarianhodis, thank you!

PR #752, opened 2016-08-11
[TREE] add a proper method to move a page as "last child of"

gobjila:
When moving a page as the last child of a parent, the called method was "moveAsFirstChildOf".
I create a new moveAsLastChildOf method to avoid confusions.

Thank you for your feedback.


gobjila:
ping @eric-chau, @crouillon thx

PR #757, opened 2016-10-14
Enabling us to change the mediafolder of a media

hbaptiste:
Modifications made in  the postAction method.


crouillon:
:+1:
Thx @hbaptiste

PR #758, opened 2016-10-17
Fix bundle overriding

crouillon:
Today BundleLoader overrides:

- configuration of bundles threw folder `/reporository/bundle/<< bundleid >>/`
- services of bundles threw folder `/reporository/bundle/<< bundle basename >>/`

This PR aims to allow services overriding using bundleid but preserves the other one


eric-chau:
nice catch @crouillon :+1: thank you

PR #759, opened 2016-11-18
fixed ExceptionListener::getErrorPageResponse to check existence of parameter before using it

eric-chau:


crouillon:
👍  nice catch @eric-chau

PR #761, opened 2016-12-14
Rename BackBee into BackBee CMS

mickaelsteinberg:
Fix URLs on Readme and rename product to BackBee CMS


crouillon:
👍 Thx @mickaelsteinberg

PR #762, opened 2016-12-23
updated AbstractContent::getData to ignore null type

eric-chau:
This PR aims to avoid error like `PHP Fatal error:  Class name must be a valid object or a string in /vendor/backbee/backbee-php/ClassContent/AbstractContent.php on line 751` which occurs if classcontent that does not exist anymore still remain in your database and `bbapp.classcontent.exception_on_unknown_classname` parameter is setted to `false`.

ping @crouillon


crouillon:
👍 Thanks @eric-chau

PR #765, opened 2017-01-27
Add new events dispatching

crouillon:
This PR introduces two new events:
* `bundle.<<bundle id>>.started` : dispatched after the bundle starts
* `bundle.<<bundle id>>.stopped` : dispatched after the bundle stops

### DIFF
--- a/Bundle/Event/BundleStopEvent.php
+++ b/Bundle/Event/BundleStopEvent.php
@@ -1,0 +1,73 @@
+<?php
+
+/*
+ * Copyright (c) 2011-2017 Lp digital system
+ *
+ * This file is part of BackBee.
+ *
+ * BackBee is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * BackBee is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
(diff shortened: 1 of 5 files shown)


Bundle/Event/BundleStopEvent.php
```diff
@@ -0,0 +1,73 @@
...
+
+use BackBee\Bundle\BundleInterface;
+use BackBee\Event\Event;
+
+/**
+ * Event dispatch on bundle stop.
+ *
+ * @author Charles Rouillon <charles.rouillon@lp-digital.fr>
+ */
+class BundleStopEvent extends Event
```
eric-chau: I think you should create an abstract class (`AbstractBundleEvent` for example) so `BundleStartEvent` and `BundleStopEvent` can extends it. Because currently `BundleStartEvent` and `BundleStopEvent` are strictly equals so it is code duplication.

Bundle/Listener/BundleListener.php
```diff
@@ -46,6 +48,11 @@ public static function onApplicationStop(Event $event)
         foreach (array_keys($container->findTaggedServiceIds('bundle')) as $bundleId) {
             if ($container->hasInstanceOf($bundleId)) {
                 $container->get($bundleId)->stop();
```
eric-chau: Maybe you can change it to limit the number of uses of `$container`:

```php
$eventDispatcher = $container->has('event.dispatcher') ? $container->get('event.dispatcher') : null;
foreach (array_keys($container->findTaggedServiceIds('bundle')) as $bundleId) {
    if ($container->hasInstanceOf($bundleId)) {
        $bundle = $container->get($bundleId);
        $bundle->stop();
        if ($eventDispatcher) {
            $eventDispatcher->dispatch(sprintf('bundle.%s.stopped', $bundleId), new BundleStopEvent($bundle));
        }
}
```

eric-chau:
:+1: nice addition @crouillon. Please take a look at my comments ;)

PR #769, opened 2017-03-01
[ClassContent] fixed RevisionRepository::getDraft to handle exception on multiple drafts for a single content

eric-chau:


crouillon:
👍 Thanks

PR #772, opened 2017-03-08
[BUGFIX] Ensure unicity of a generated name for an uploaded file.

crouillon:
Uploading two images with the same original name in the same second generates one same filename.
So only one image is saved.

This patch replaces the descriminator time() by uniqid ('', true) to ensure the unicity of the filename.


fkroockmann:
:+1:

PR #773, opened 2017-03-23
[Rest] updated SecurityController to encode password with declared encoder

eric-chau:
Currently, `SecurityController::authenticateAction` only encodes password with `md5`. This PR aims to change it by using the declared encoder for `BackBee\Security\User` in `security.yml`.


crouillon:
👍 Nice catch @eric-chau

PR #775, opened 2017-03-27
[Rest] updated ResourceController to dispatch ValidateFileUploadEvent

eric-chau:
This PR adds a new event: `ValidateFileUploadEvent`. This event is dispatched when a new base64 encoded file is uploaded. By default, BackBee accepts the upload of any files. In some cases, you may want to be able to restrict the upload of images only.


crouillon:
:+1:

PR #779, opened 2017-03-30
[Event] fixed PageListener to use count() instead of iterator->count()

eric-chau:
In some cases, `$notDeletedDescendants` is type of array so the method `::getItrator()` does not exist.


crouillon:
Thx @eric-chau 
:+1:

PR #781, opened 2017-04-03
[BUGFIX] [Renderer] Fix insert new item in array at any position

crouillon:
While adding new layouts, helpers and scripts dir at default position 1, key mismatch and overriding occured


eric-chau:
👍 nice catch @crouillon, thank you =)

PR #783, opened 2017-04-27
Catch SecurityException in AuthenticationManager

crouillon:
Because a SecurityException thrown by providers was not catch, the authentification process ended before the last authentication providers was called, disabling the possible providers defined after the core ones.


eric-chau:
:+1: thank you @crouillon

PR #785, opened 2017-05-29
[MISC] Remove support of PHP < 5.6

crouillon:
Resolved [#784](https://github.com/backbee/backbee-php/issues/784)
Due to an update of doctrine/common, php < 5.6 can no more be supported

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,6 @@ language: php
 sudo: false
 
 php:
-    - 5.4
-    - 5.5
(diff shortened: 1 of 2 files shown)


eric-chau:
:+1: should we update PHP minor version? `1.3.*`? @crouillon. This PR aims to solve issue #784.

crouillon:
Thanks @eric-chau, I agree, we should update minor version on 1.3.0 for the four projects

Issue #786, opened 2017-06-02
Page publication date


cedric-b:
The page publication date should be set to the date of the first status change to online if not explicitly set through the publication/depublication programmer.
Without this, there's no page publication date when not explicitly set and the only reliable date is the page creation date, which is not very accurate/pertinent.

PR #787, opened 2017-06-02
[#786] Page publication date

cedric-b:
Resolve the issue in reference

### DIFF
--- a/NestedNode/Page.php
+++ b/NestedNode/Page.php
@@ -950,6 +950,10 @@ class Page extends AbstractObjectIdentifiable implements RenderableInterface, Do
             throw new \LogicException("Root page state must be online.");
         }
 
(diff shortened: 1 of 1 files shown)


NestedNode/Page.php
```diff
@@ -950,6 +950,10 @@ public function setState($state)
             throw new \LogicException("Root page state must be online.");
         }
 
+        if (self::STATE_OFFLINE === $this->_state && self::STATE_ONLINE === $state && !$this->getPublishing()) {
```
crouillon: Should be something like:
if (!$this->isOnline() && (self::STATE_ONLINE & $state) &&  !$this->getPublishing()) {

crouillon:
:+1: thanks @cedric-b

PR #789, opened 2017-06-15
decreased ContainerListener::onApplicationInit priority to ensure that container dump is complete

eric-chau:


crouillon:
👍 hanks @eric-chau

PR #791, opened 2017-06-19
[Doctrine] added RepositoryFactory to ease customization of entity repository

eric-chau:
…


crouillon:
:+1: Thanks @eric-chau

Issue #792, opened 2017-06-28
Authentication error when password is not right.


Bensid:
`Argument 2 passed to Symfony\Component\Security\Core\Event\AuthenticationFailureEvent::__construct() must be an instance of Symfony\Component\Security\Core\Exception\AuthenticationException, instance of BackBee\Security\Exception\SecurityException given, called in /var/www/backbee-cms/vendor/backbee/backbee-php/Security/Authentication/AuthenticationManager.php on line 136 and defined`

PR #793, opened 2017-06-28
[Security] fixed error 500 which is raised on wrong password #792

eric-chau:
This PR aims to fix issue #792.


eric-chau:
ping @crouillon =)

crouillon:
:+1: Thanks @eric-chau

PR #794, opened 2017-07-04
Add classcontent.streamparsing event dispatching

crouillon:
This new event is triggered just after the YAML file parsing and just before the PHP class build.

It gets a read/write `data` argument containing the array representation of the parsed yaml.


eric-chau:
:+1: thank you @crouillon

crouillon:
Thanks for review

PR #796, opened 2017-08-07
[#795] - [Authenticate] - Fix headers for Chrome 60

Bensid:


Bensid:
Fix headers for authentication on Chrome 60

PR #798, opened 2017-09-22
[Event] added Dispatcher::clearAllListeners() to give proper way to clear all listeners

eric-chau:


crouillon:
:+1: tks @eric-chau

PR #802, opened 2017-11-29
[Rest] fixed SecurityListener to not throw exception on null token

eric-chau:


crouillon:
:+1:
Thanks @eric-chau

PR #806, opened 2018-02-15
[BugFix] Avoid strict standard warning while counting a non-array variable

crouillon:


eric-chau:
:+1:

crouillon:
Thanks @eric-chau

PR #807, opened 2018-02-16
Insures the name of the altered table is quoted according to the platform

crouillon:


eric-chau:
:+1: thank you @crouillon

crouillon:
thx @eric-chau

PR #808, opened 2018-07-02
[ClassContent] fixed Revision::commit() to not use undefined variable

eric-chau:


crouillon:
Nice catch, thank you @eric-chau 
👍

